PR #43, opened 2024-09-13
[Emus] Add chimerasnes as default emulator for snes

cobaltgit:
This pull request adds an optimised build of the ChimeraSNES libretro core as the new default Super Nintendo emulator, with better performance* than the `mednafen_supafaust` core and RetroAchievements compatibility!

\**Super Mario World 2: Yoshi's Island*'s title screen runs at 60fps with CPU clock set to 1200MHz and SuperFX clock set to 10MHz


ryanmsartor (approved):
Looks good! Thanks for working to make spruce an even better experience for everyone!

PR #47, opened 2024-09-18
Development

decojon:
Files modified to make the hybrid sleep function work and for the power button to be used for quick-save.


ryanmsartor:
We will only be pulling directly from RC to main immediately following a new version release from now on. New features should be developed on their own branch and a pull request made from that branch to RC after testing has been done.

PR #51, opened 2024-09-19
Automatic Icon refresh after theme change

XanXic:
Makes a check that when returning to the menu if their theme changed refresh the icons


fengfrw:
It is not suitable for automatic refresh for icon packages that want to keep using other themes.

Issue #107, opened 2024-10-15
set uae4arm default for Amiga, add Atari ST folder for games


metallic77:
As the title says the uae4arm core is way faster than PUAE for Amiga games.  It runs most games full speed even on my raspberry pi 2w, even heavy copper use ones like "Lionheart". Also add a folder for Atari ST games (Hatari default), i haven't seen one, there is an ATARI which i guess is for 2600

cobaltgit:
This pull request should do the job. Yes, the ATARI folder is for 2600 games.

metallic77:
That was quick, thanks. Lakka that is linux and retroarch works wonders on pi 2w with uae4arm. Haven't tested this much but it should perform the same.

cobaltgit:
#108 merged

cobaltgit:
I wonder why uae4arm is always running at half speed without fastforward...

metallic77:
I noticed the raspberry pi lakka version has an option for fast copper too, that's missing from this version. Uae4arm should be full speed normally.

cobaltgit:
Hm, I've just compiled uae4arm libretro from source and it has the option

metallic77:
Good, how does it perform?

cobaltgit:
A bit better in fastforward mode (i compiled with cortex-a7 tune) but more or less the same, I can't nail down why it's running at half speed without fastforward

cobaltgit:
Atari ST seems to boot into the Hatari menu for some reason

metallic77:
It works if launched from retroarch but not launching games from main menu, i transferred your changes to sd card.

metallic77:
> A bit better in fastforward mode (i compiled with cortex-a7 tune) but more or less the same, I can't nail down why it's running at half speed without fastforward

I have some clues. First of all core runs at 60 fps meaning NTSC, while most amiga games made in pal 50hz. Second if you disable audio sync in settings-audio then bang, solid 60 fps at all times. There is a) format is wrong, should be pal and b) audio sync

cobaltgit:
Thanks for the tip...pushed the optimised core and config override to the devel branch ([1e8f1b9](https://github.com/spruceUI/spruceOS/commit/1e8f1b965954ec33e15f0d8a2d08c9adb24e464b))

metallic77:
[...]

PR #108, opened 2024-10-15
Add Atari ST emu folder and change Amiga core to uae4arm

cobaltgit:
Changing the Amiga core should improve performance, this all may need testing. Addresses issue #107

### DIFF
--- a/Emu/.emu_setup/core/puae.sh
+++ b/Emu/.emu_setup/core/puae.sh
@@ -4,6 +4,6 @@ EMU_NAME="$(echo "$1" | cut -d'/' -f5)"
 CONFIG="/mnt/SDCARD/Emu/${EMU_NAME}/config.json"
 SYS_OPT="/mnt/SDCARD/Emu/.emu_setup/options/${EMU_NAME}.opt"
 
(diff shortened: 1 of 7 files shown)


cobaltgit:
Whoops, didn't mean to assign myself

metallic77:
I did some testing, uae4arm core here doesn't perform the same as lakka uae4arm, either because it's 64bit there, or they are different versions with many changes or some bug here because if i ffwd it runs smooth (without frame skipping on ffwd). While puae here performs a lot better better than puae on pi 2w lakka.

cobaltgit:
I've done some testing myself, the uae4arm core seems to have the best performance out of the three available Amiga cores on spruce, with puae2021 being a close second and puae (current) being the worst overall.

Maybe it's because the zero 2 w (at least when overclocked) has better cpu performance (a53 64bit vs a7 32bit)

metallic77:
Yes i would avoid the puae current for this device.  uae4arm runs at half speed, meaning about 25 fps on my a30 (updated firmware to 0713). But when i ffwd it seems like it's capable for more.

cobaltgit:
I'll take a closer look tomorrow, I noticed that too.

PR #110, opened 2024-10-15
Feat:  Add Syncthing Sync Check 

BaseInfinity:
### Problem

There are plenty of times where I would save the game, and then power off the device and the Syncthing App would not be done uploading the local saves to the remote device.

This is a pain because it requires me to save the game, go to menu, look at the Web GUI and wait for the device to finish.

This can take sometimes 10-30 Seconds depending on how big the save state is, how long it takes for Syncthing to trigger a resync, and how good ur interwebs is.

The Inverse would happen too when I would auto resume a game but syncthing wasn't finished downloading the remote changes yet.

If using Auto Save/Loads, this can be even more problematic as you will clobber your saves and create a mess

### Solution

Use the Syncthing API **only when we need to** know when to Force a sync check and poll for completion on either Startup or Shutdown phases of the device.

This means when starting up a game, ensuring all remote files are done downloading to A30.
This means when shutting down the device, ensuring all A30 files are done uploading to remote device.

### Places I've inserted the Sync Check:
* Hold Power Button to Quick Save + Shutdown
* Power On and jump into the AutoResumeRA script
* If you're launching an emulator for the first time during a session

Every other situation I felt the device didn't need to be aggressive about syncing and these were the 3 major pain points

### Considerations
* Had to move starting background services lower in the `runtime.sh` so we jump into the game ASAP if AutoResume is enabled. Otherwise this adds considerable time and slows down the overall syncthing status and bootup time
* The User Experience can be jarring, every 3 seconds i update the status of both the folders in sync using the display function, this might be too aggressive or jarring or too much stuff to show
* I used a temporary file for checking/reading the sync status because of the limitation of subshells, but I just realized I could have used global variables instead but oh well, it's fast and barely runs
* Weird stuff happens ATM with the `pwrbtn.sh` functionality atm that I believe is not tied to my code. I used to have it rigged up to the `select.sh` script and had no issues

It's probably best to go commit by commit to get an understanding of why i put things where and I should have a Demo Video once I can get a stable Dev build goin


App/Syncthing/syncthing_sync_check.sh
```diff
@@ -0,0 +1,284 @@
...
+    # Set the API key
+    if ! set_api_key; then
+        display -t "Error: Unable to find API key"
+        sleep 2
+        stop_network
+        exit 1
+    fi
+
+    case "$1" in
+        --monitor)
```
BaseInfinity, line 247: This is only used for testing purposes. There were times I just wanted a constant stream of status while SSH'd into the device

Emu/.emu_setup/standard_launch.sh
```diff
@@ -62,6 +62,16 @@ if [ "$MODE" != "overclock" ] && [ "$MODE" != "performance" ]; then
 	/mnt/SDCARD/spruce/scripts/enforceSmartCPU.sh &
 fi
 
+##### SyncThing Sync Check, perform only once per session #####
```
sterlinghirsh: Let's standardize on one capitalization of syncthing. Pretty much everywhere else it's Syncthing.

App/Syncthing/syncthing_sync_check.sh
sterlinghirsh: This would be a good code comment

App/Syncthing/syncthing_sync_check.sh
```diff
@@ -0,0 +1,284 @@
...
+Press START to cancel"
+
+    if [ "$mode" = "shutdown" ]; then
+        force_rescan
+        sleep 1
+    elif [ "$mode" = "startup" ]; then
+        force_rediscovery
+        # THIS MIGHT NOT BE NEEDED, WE MIGHT BE ABLE TO JUST SLEEP 1
+        # sleep 1
+        sleep 5
```
sterlinghirsh: If it's not needed, it seems better to sleep 1 than 5.

App/Syncthing/syncthing_sync_check.sh
```diff
@@ -0,0 +1,284 @@
...
+
+. /mnt/SDCARD/App/Syncthing/syncthingFunctions.sh
+. /mnt/SDCARD/spruce/scripts/helperFunctions.sh
+
+SYNCTHING_CONFIG_DIR="${SYNCTHING_CONFIG_DIR:-/mnt/SDCARD/App/Syncthing/config}"
+API_ENDPOINT="http://localhost:8384/rest"
+CONFIG_XML="$SYNCTHING_CONFIG_DIR/config.xml"
+CHECK_INTERVAL=2  # Interval between checks in seconds
+MAX_API_RETRIES=15
+API_RETRY_INTERVAL=2
```
sterlinghirsh: Any reason not to make these intervals 1? They seem like quick checks that take much less than 1 second, but not sure if you found 2 seconds works better for some reason.

spruce/scripts/autoRA.sh
```diff
@@ -1,36 +1,42 @@
...
-	done	
-}
-
 if flag_check "save_active"; then
 	log_message "Save active flag detected"
+
+	wifi_needed=false
+	syncthing_enabled=false
+
+	keymon &
```
sterlinghirsh: What is this for?

App/Syncthing/syncthing_sync_check.sh
```diff
@@ -0,0 +1,284 @@
...
+            log_message "Sync check cancelled by user"
+            return 1
+        fi
+
+        # Remove the status files at the beginning of each iteration
+        rm -f /tmp/sync_status
+        rm -f /tmp/sync_display.txt
+    
+        # TODO: Consider refactoring to use in-memory approach instead of temporary files
+        # for performance improvement and reduced I/O. Current file-based method:
```
sterlinghirsh: Is this spending a considerable amount of time on file io? I'd be surprised if the performance was meaningfully different between a file and memory unless you're reading or writing to it a lot.

App/Syncthing/syncthing_sync_check.sh
```diff
@@ -0,0 +1,284 @@
...
+        return 1
+    fi
+
+    while true; do
+        local all_synced=true
+        local summary=""
+        local status_lines=""
+
+        # Check if the Start button was pressed
+        if tail -n1 "/var/log/messages" | grep -q "enter_pressed 0"; then
```
sterlinghirsh, line 154: Is it a reasonable concern that a different button would be the last one pressed between checks? Is every button press getting logged to messages?

App/Syncthing/syncthing_sync_check.sh
```diff
@@ -0,0 +1,284 @@
...
+    fi
+}
+
+stop_network() {
+    log_message "Stopping network interface..."
+    ifconfig lo down
+}
+
+get_folders() {
+    curl -s -H "X-API-Key: $API_KEY" "$API_ENDPOINT/config/folders" | jq -r '.[] | "\(.id)|\(.label)"'
```
sterlinghirsh, line 59: Consider a function or even just a string that encapsulates `curl -s -H "X-API-Key: $API_KEY" "$API_ENDPOINT"` since it gets repeated a lot of times.

sterlinghirsh:
This looks pretty good! CR :zap:

App/Syncthing/syncthing_sync_check.sh
ryanmsartor: correct! All button presses get logged to that messages file - and even non-button presses. (This is something baked into Miyoo's closed source firmware which we are building upon and fighting against)

App/Syncthing/syncthing_sync_check.sh
BaseInfinity: That's a good point, the first `CHECK_INTERVAL` is set at 2 seconds because in the same function I update the display and doing it every second could be jarring and not give the person time to read.

The `API_RETRY_INTERVAL` isn't tied to displaying any information so I could poll quicker, gracias:

https://github.com/spruceUI/spruceOS/pull/110/commits/0e0079269c035a3aa88f52753e3204e0b20dc769

App/Syncthing/syncthing_sync_check.sh
BaseInfinity: I'm okay with leaving this as-is, i feel it's more readable this way even though it's a bit repetitive

App/Syncthing/syncthing_sync_check.sh
BaseInfinity: Definitely, I wanted to do another round of testing with this sleep changed and see if it's enough.

This originally was set to 5 seconds because I was calling the force_rediscovery endpoint correctly. I'll do more testing and report back (I definitely wont merge this without that test)

App/Syncthing/syncthing_sync_check.sh
BaseInfinity: You're right, i think I was exaggerating here on potential resource usage. Ill remove this TODO:

https://github.com/spruceUI/spruceOS/commit/f5fdab1d534e98d6956c40e79a7fb85695a4d390

App/Syncthing/syncthing_sync_check.sh
BaseInfinity: Good call!

https://github.com/spruceUI/spruceOS/pull/110/commits/cddc54f8e205d76fcc53086559910a5e6a685000

Emu/.emu_setup/standard_launch.sh
BaseInfinity: Lol fairrrrr, while building this feature, the first half of the time I thought this was SyncThing, whoopsies:


https://github.com/spruceUI/spruceOS/pull/110/commits/e6c3a1e00dcd746793baacc92ebecb7b207cd333

spruce/scripts/autoRA.sh
BaseInfinity: `keymon` is what monitors inputs and sends it to `/var/log/messages` 

It's also a type of pokemon

BaseInfinity:
I'm going to do some testing on this:
https://github.com/spruceUI/spruceOS/pull/110#discussion_r1803986137

After my testing I'll decide to leave the 5 seconds or change to 1.

App/Syncthing/syncthing_sync_check.sh
BaseInfinity: I did some testing and 2 seconds was the magic number. 1 second worked most of the time but 2 seconds worked all the time.
https://github.com/spruceUI/spruceOS/pull/110/commits/807be37a3ebd5ae13b0907f761e95c12fe247497

BaseInfinity:
Okay this pull request is in a place where it is working. 

Any weirdness atm is due to `pwrbtn.sh` issues.

I can do some demo vids tomorrow but If no one has any objections to merging this, we probably should do it soonish to avoid the pull request from getting stale.

BaseInfinity:
The only bummer about this which I have no control over is the check for wifi on startup.

It takes the device a solid 10+ seconds to check for Wifi when going into Auto Resume from a "Quick Save + Shutdown" path. 

This also happens currently if RetroAchievements is enabled so I guess I'm not making it any worse than it already is

BaseInfinity:
The reason that Wifi takes 10 seconds is because it actually waits for an assigned IP address, not just that the wireless adapter is started.

This can be deceiving when we display the wifi logo to the user, because they don't have an actual assigned IP address yet but meh nbd

BaseInfinity:
I need to update these changes to work with the new `powerbutton_watchdog.sh` script that is more reliable

spruce/scripts/save_poweroff.sh
```diff
@@ -83,6 +83,16 @@ show_image "/mnt/SDCARD/.tmp_update/res/save.png"
...
+if flag_check "syncthing"; then
+	log_message "Syncthing is enabled, WiFi connection needed"
+
+	if check_and_connect_wifi; then
+		/mnt/SDCARD/App/Syncthing/syncthing_sync_check.sh --shutdown
+	fi
+
+    flag_remove "syncthing_startup_synced"
+fi
+
```
BaseInfinity, line 95: The code was deleted in the previous merge conflict, so while it looks added here, it actually moved from the powerbutton watchdog script

PR #113, opened 2024-10-16
add simple light shaders, compatible to GLES 2.0

metallic77:
They work fine on my A30, full speed easily. They can resolve the nasty scaling when full screen (without resolution being an exact multiply). With these you can occupy all screen on gba, gb, gbc, gg, snes without nasty bad scalling artifacts

### DIFF
--- a/RetroArch/.retroarch/shaders/gritty.glslp
+++ b/RetroArch/.retroarch/shaders/gritty.glslp
@@ -1,0 +1,5 @@
+shaders = "1"
+
+feedback_pass = "0"
+shader0 = "shaders_glsl/gritty.glsl"
+filter_linear0 = "true"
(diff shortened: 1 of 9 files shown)


ryanmsartor:
Nice! May I inquire about the folder structure? Is there a reason it needs to be nested the way it is? I'd like to have all the presets in one flat folder if possible, for the easiest time for new users to find and select among them, especially since you're not adding so many that the extra folders would make a significant improvement in organization.

metallic77:
It follows retroarch folder structure, example q preset is a glslp file and it points to a folder that has the actual glsl shader. It's better organised this way imo

metallic77:
Reverting with all presets in one place.

PR #115, opened 2024-10-16
pico8 vibrate on exit when long press power button

cinethe-zs:
pwr button long press to exit and import.sh only import cart game file


ryanmsartor (approved):
Looks good!

PR #126, opened 2024-10-23
replace old hatari-fuse with latest builds

metallic77:
Don't know if this is the proper way, if not someone grab the core test it and do it ;)

compiled in MX Linux and docker with instructions here
https://github.com/zoltanvb/retroarch-cross-compile

Hatari 1.8 7008194
Large update with many options changed, previous core was too old, like 2 years old  
cpu speed needs to be 1200 for full speed in my tests 

fuse ZX Spectrum 1.6.0 dc116e6
Adds speed options on core options (50-100-150-200-300) and ability to start without a game, also some changes in controllers. Definitely worth the update.


cobaltgit (changes requested):
The new fuse core runs nicely, from my testing with smart CPU mode it runs at 1008mhz. As for hatari, I need to test that with a non-IPF ROM as there's no support for it. Could you perhaps build the core with CAPS support for IPF ROMs (requires `libcapsimg.so` in the BIOS directory), and in the meantime, I'll test with a non-IPF ROM. Thank you in advance, all's good so far :)

cobaltgit:
On another note, I'm currently testing an STX game and it works well! Running at 1344mhz in smart mode

metallic77:
Isn't stx (pasti) what most atari emulators used for copyrighted disks? Was the previous hatari loading ipf files?

cobaltgit:
> Isn't stx (pasti) what most atari emulators used for copyrighted disks? Was the previous hatari loading ipf files?
No, I don't think that was compiled with caps support either, gotcha

> that enables it, you need a compiled caps image for the architecture too. RetroArch isn't distributing the Hatari core with IPF support files due to GPL license as far as i know
Aha, just seen capsimg is licenced under the MAME licence.

metallic77:
There is a PR not added 

https://github.com/libretro/hatari/pull/55/files

I could try to merge the PR separately and see if it compiles well

metallic77:
Try this core and tell me if it works

https://github.com/metallic77/shaders_glsl-slang/blob/main/hatari_libretro_ipf.so

Cloned and compiled the guy's that made the PR repository, still older than current retroarch hatari

cobaltgit:
In all fairness I'm not sure we need ipf support yet, we could add it later down the line seeming as the previous core didn't support it so it isn't really a downgrade.

cobaltgit (approved):

PR #143, opened 2024-10-29
setting app now can show ip & mac addresses

oscarkcau:


oscarkcau:
out dated so close the request

PR #178, opened 2024-11-09
Add GS global toggle

cobaltgit:
WIP


cobaltgit:
Seems to be passing all my tests so far

PR #181, opened 2024-11-11
fix: Syncthing check will now fail gracefully if no folders are configured

BaseInfinity:
Syncthing would be forever syncing or freeze if no folders were found but a device was connected.

Now we check for folders first and gracefully exit


BaseInfinity:
Video of it working:
https://discord.com/channels/1252066540874891355/1305288827841282140/1305605442080211035

BaseInfinity:
Note to self:

Now that we have GH Actions going, I could technically setup regression unit/integration tests for Syncthing but it is definitely overkill at this time.

PR #223, opened 2024-11-26
ST Sync Check:  Make faster and more robust

BaseInfinity:
### Problem
ST Sync Check was using long static sleeps and always assumed worse case scenario. This meant punishing everyone even if they kept their Wifi and Network services enabled in game.

Also due to how we kill network services in game, we sometimes put ST in a broken state when wifi/wlan would get killed before ST shuts down

### Solution
* Ensure we kill network services in the foreground so ST exits gracefully and isn't put in a broken state due to a race condition with wifi/wlan
* Use polling mechanisms to ensure that script works for all types of possible toggles, such as `Disable Network Services in Game` and `Disable Wifi in Game`

### DIFF
--- a/Emu/.emu_setup/standard_launch.sh
+++ b/Emu/.emu_setup/standard_launch.sh
@@ -82,11 +82,12 @@ if $syncthing_enabled && $wifi_connected; then
     start_syncthing_process
     /mnt/SDCARD/spruce/bin/Syncthing/syncthing_sync_check.sh --startup
     flag_add "syncthing_startup_synced"
+
 fi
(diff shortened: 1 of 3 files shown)


BaseInfinity:
I tested the crap out of this and have a ton of working videos testing all the state permutations on Discord such as:

Disable Network Services in Game
Disable Wifi Services in Game

Toggle one on, toggle one off etc. 

imma merge this so it's in 3.1.0, it fixes bugs around ST regarding 3.0.1 and performs like 80% faster across the board

PR #234, opened 2024-11-30
feat(multipass): add marking hidden network in multipass

Gerard-Szulc:
Adds possibility to flag multipass network if it is hidden.


cobaltgit:
What does this do? Also you're merging into the wrong branch, open it into Development instead of main

Gerard-Szulc:
Sorry, I missclicked branch. Fixing it and I'll add more info

PR #235, opened 2024-11-30
feat(multipass): add marking hidden network in multipass

Gerard-Szulc:
Allows user to tell if network specified in multipass file is hidden by providing "1" value.

I added reading HIDDEN_1, HIDDEN_2 ...... from multipass and according to its value scan_ssid=1 is added in wpa_supplicant

According to https://wiki.archlinux.org/title/Wpa_supplicant#Connection if the network is hidden, wpa_supplicant should have scan_ssid=1.

### DIFF
--- a/multipass.cfg
+++ b/multipass.cfg
@@ -18,15 +18,20 @@
 
 ID_1=""
 PW_1=""
+HIDDEN_1=""
 
 ID_2=""
 PW_2=""
+HIDDEN_2=""
 
 ID_3=""
 PW_3=""
+HIDDEN_3=""
 
 ID_4=""
 PW_4=""
+HIDDEN_4=""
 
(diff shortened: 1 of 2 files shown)


ryanmsartor:
closed by accident when I accidentally deleted the Development branch ahaha

ryanmsartor:
@Gerard-Szulc I like what you did - may I just make one request before I merge this?

Can you please add instructions for users on how to use this new field, to the comments at the top of multipass.cfg?

Gerard-Szulc:
Sorry, didnt have time last week, I'll try to add this tomorrow, and there is one thing that I need to check addidtionally. Also I probably would need to buy new a30 soon, because i cant get one(we have 2 a30) from my kids to test everything properly :P

PR #273, opened 2024-12-13
DRAFT: New box art scrapper based on ScreenScraper.fr

njgreb:
Initial take on a new box art screen scraper.

A few obvious issues:
1) Doesn't support all systems yet
2) No clue how this compares in speed to the previous version (I just got my A30....so I've never seen the other version work)
3) Depends on a single service
4) Um...am I supposed to be sharing these credentials like this...the people behind ScreenScraper.fr just don't respond fast.

Possible paths forward:
1) Just add more systems :)
2) We could add some caching layer, but that is just a new dependency
3) I can make this work with both once libretro is back online
4) Waiting on an answer from them

There is still plenty of work to do here, but mostly wanted to get thoughts on the approach and issue #2.


XanXic:
Hey, thanks for taking a crack at this. We were focused on getting the 3.2.0 release out but we are done. So could you update this to the latest version? Right now this is adding back all the images we eliminated lol.

Ie the ApiKey, yeah I really doubt we'd want to keep that. Is that your personal development key in there? I'm not familiar with screen scrapers api key system. I know they will restrict requests much harsher without a login.

I would like to figure out screen scrapers process more, but I think initially this needs an easy way to let someone pass in their own login/api credentials. Right now they'd have to edit that line which is more than a lot of users could do. But even having like a 'credentials' file with "username=", "password=" that they can easily edit and is read from by the script would go a long way to making it easy for users to login. If we kept it as is.

PR #290, opened 2024-12-22
add context option to "Add to Favorites" within the x button menu

Kickbut101:
"add_favourite.sh" - Quick script that allows you to pass in the game name and rom directory (from standard launch.sh) and add those parameters to the favourite.json file

Additionally added the context menu option to each of the emulators .config file.


Kickbut101:
Currently this is a sort of slow method. Since it utilizes the launch.sh script to make it happen. I'm not versed enough to make it happen faster or "first class" like the home button does. But that ideally would be the next step. Hopefully someone can take it a bit further and make this nicer/better?

Sundownersport:
Hey @Kickbut101 Sorry this kind of sat around for a while.

How is this coming along? I wonder about getting it added soon here.

Kickbut101:
> Hey @Kickbut101 Sorry this kind of sat around for a while.
> 
> How is this coming along? I wonder about getting it added soon here.

Was there something further for me to do?

It's in a state AFAIK that should be ready to test and merge?

Emu/.emu_setup/add_favourite.sh
```diff
@@ -0,0 +1,23 @@
...
+EMU_NAME="$(echo "$1" | cut -d'/' -f5)"
+ROMFILENAME="$(basename "$1")"
+GAMENAME="$(echo "$ROMFILENAME" | cut -d'.' -f1)"
+
+log_message "--- Adding favorite game chosen by user ---"
+
+# Construct the new JSON entry
+NEW_ENTRY=$(jq -n \
+  --arg label "$GAMENAME ($EMU_NAME)" \
+  --arg launch "/mnt/SDCARD/Emu/$EMU_NAME/launch.sh" \
```
cobaltgit, line 17: We have removed individual emu launch scripts in favour of directly running `.emu_setup/stanard_launch.sh`

Emu/.emu_setup/add_favourite.sh
```diff
@@ -0,0 +1,23 @@
...
+  --arg rompath "/mnt/SDCARD/Emu/$EMU_NAME/../../Roms/$EMU_NAME/$ROMFILENAME" \
+  --argjson type "5" \
+  '{label: $label, launch: $launch, rompath: $rompath, type: $type}')
+
+# Append new json formatted entry to favourite file
+echo "$NEW_ENTRY" >> $FAVOURITE_FILE
```
cobaltgit, line 23: Maybe we could just run the jq command and output to the file rather than storing as a variable than echoing? Like with the last comment I could always modify it after the PR is merged

cobaltgit (commented):
Nice script, I myself don't know how to add to favourites via MainUI for whatever reason ahah

PR #331, opened 2025-01-24
feat(homebutton): first pieces of homebutton_watchdog for the Brick

dhoizner:
First pieces of enabling the home button watchdog on the Brick.

### Currently:
- [x] kills current app on home+start
- [x] kills retroarch on home+start
- [x] opens in-game menu in retroarch

### Todo:
- [ ] Pause in RA needs to be fixed
- [ ] PPSSPP
- [ ] Drastic
- [ ] Other home actions


cobaltgit:
Should also work with the smart pro if we add the appropriate platform checks

dhoizner:
merging, we can clean up whatever breaks when we find it

PR #379, opened 2025-04-13
Flip: Use 64-bit bins, fix led path, add platform checks

codexblack:
Have been testing out Spruce on my Miyoo Flip V2. There is still a lot I don't understand about this project but these changes make sense to me and I can't see anything that breaks after.


spruce/scripts/save_poweroff.sh
```diff
@@ -113,6 +117,15 @@ if pgrep "ra32.miyoo" >/dev/null; then
     # } | $BIN_PATH/sendevent /dev/input/event3
     # sleep 0.3
     killall -q -15 ra32.miyoo
+elif pgrep "ra64.miyoo" >/dev/null; then
+    # {
```
codexblack, line 117: This is just copied from ra32.miyoo

spruce/settings/platform/Flip.cfg
```diff
@@ -5,4 +5,4 @@ export DISPLAY_WIDTH="640"
 export DISPLAY_HEIGHT="480"
 export SYSTEM_JSON="/userdata/system.json"
 export BATTERY="/sys/class/power_supply/battery"
-export LED_PATH="/sys/class/leds/work"
```
codexblack: This broke the LEDs on my Flip

PR #434, opened 2025-04-19
Use the new flip libmali for all portmaster games

chrisj951:
Might revert and do it for JUST games using weston, but doing it for all for now

Unmount after port.

Also testing if we just want this on boot for flip vs just for ports


chrisj951:
Found a few unworking weston games that this fixes, so leaving for at least all PM games

Defshot
ROTA

PR #437, opened 2025-04-20
Start moving emufresh into python for the flip

chrisj951:
Left a few parts of it in the bash script


spruce/scripts/emufresh_md5_multi.sh
```diff
@@ -2,286 +2,306 @@
...
-	# hide all systems
-	find "$roms_path" -mindepth 1 -maxdepth 1 -type d | while read -r folder; do
-		system_name=$(basename "$folder")
-		config_file="$emu_path/$system_name/config.json"
-		sed -i 's/^{*$/{{/' "$config_file"
-	done
-	if [ -d "$roms2_path" ]; then
-		find "$roms2_path" -mindepth 1 -maxdepth 1 -type d | while read -r folder; do
+prompt_main_ui_restart() {
+	echo "prompt_main_ui_restart()"
```
dhoizner: should this be a `log_message`?

chrisj951: Ive removed it in the commit of adding python, was there for debugging only

PR #438, opened 2025-04-20
Fix bluetooth for Flip

00svd00:
Only works for latest official firmware (miyoo355_fw_20250228)
If Gamma Loader was flashed earlier - official firmware should be reflashed.

### DIFF
--- a/spruce/scripts/principal.sh
+++ b/spruce/scripts/principal.sh
@@ -184,6 +184,9 @@ while [ 1 ]; do
                 ;;
             "Flip" )
                 export LD_LIBRARY_PATH=/usr/miyoo/lib:$LD_LIBRARY_PATH
+                insmod /lib/modules/rtk_btusb.ko
+                runifnecessary "btmanager" /usr/miyoo/bin/btmanager
+                runifnecessary "hardwareservice" /usr/miyoo/bin/hardwareservice
                 runifnecessary "miyoo_inputd" /usr/miyoo/bin/miyoo_inputd
                 cd /usr/miyoo/bin/
                 MainUI
--- a/spruce/scripts/runtime.sh
+++ b/spruce/scripts/runtime.sh
@@ -102,6 +102,8 @@ elif [ "$PLATFORM" = "Flip" ]; then
         mkdir -p /mnt/sdcard/Saves/userdata-flip/lib/bluetooth
     fi
     mount --bind /mnt/sdcard/Saves/userdata-flip/ /userdata
+    mkdir -p /run/bluetooth_fix
+    mount --bind /run/bluetooth_fix /userdata/bluetooth
 fi
 
 # Flag cleanup


Sundownersport:
Thanks for the PR! 

Are you on Discord? The team would love to chat more about development

00svd00:
Yes, I have discord account, but I barely use it
00svd00_73549

spruce/scripts/runtime.sh
```diff
@@ -102,6 +102,8 @@ elif [ "$PLATFORM" = "Flip" ]; then
         mkdir -p /mnt/sdcard/Saves/userdata-flip/lib/bluetooth
     fi
     mount --bind /mnt/sdcard/Saves/userdata-flip/ /userdata
+    mkdir -p /run/bluetooth_fix
+    mount --bind /run/bluetooth_fix /userdata/bluetooth
```
chrisj951, line 106: this is done 10 lines above.

We make /mnt/sdcard/Saves/userdata-flip/bluetooth and then mount 
/mnt/sdcard/Saves/userdata-flip/ over /userdata

chrisj951 (changes requested):

spruce/scripts/runtime.sh
00svd00: No, it doesn't work
bluetooth service tries to create folder, named as BT MAC address(like AA:BB:CC:11:22:33), and such name is not possible for fat32 partition
We also don't need to preserve /userdata/bluetooth folder content between reboots, because it's only cache there, so tmpfs works perfectly well

spruce/scripts/principal.sh
```diff
@@ -184,6 +184,9 @@ while [ 1 ]; do
                 ;;
             "Flip" )
                 export LD_LIBRARY_PATH=/usr/miyoo/lib:$LD_LIBRARY_PATH
+                insmod /lib/modules/rtk_btusb.ko
+                runifnecessary "btmanager" /usr/miyoo/bin/btmanager
+                runifnecessary "hardwareservice" /usr/miyoo/bin/hardwareservice
```
chrisj951, line 189: This change is great though, it does fix bluetooth on the flip when running spruce except it causes my flip to reboot after about 1-2min.

chrisj951: This might be something wrong with my flip actually

00svd00: Can't reproduce this. Played GBA and PSP for 10 minutes - no problems at all. But if I'll revert patches made to runtime.sh - things become crazy. Services start to crash, retroaarch barely loads, etc.

spruce/scripts/runtime.sh
chrisj951: We've had issues of the internal partitions corrupting requiring fsck to get working again.

Do you know if that folder is necessary? BT seems to work without it

spruce/scripts/principal.sh
chrisj951: Reset all my other changes w/ just this and theres 0 issues now for this part

spruce/scripts/runtime.sh
00svd00: I don't really sure. I see that without it bluetoothd spams errors to stdout.
But as I know, tmpfs should be pretty safe to write to it. It's located in ram so FS errors should never ever happen.

chrisj951: Ah if /run is tmpfs im good with it then

chrisj951 (approved):

spruce/scripts/runtime.sh
00svd00: Yeah. All of them
```
mount | grep tmpfs
devtmpfs on /dev type devtmpfs (rw,relatime,size=477572k,nr_inodes=119393,mode=755)
tmpfs on /tmp type tmpfs (rw,relatime,size=489124k,nr_inodes=122281)
tmpfs on /run type tmpfs (rw,nosuid,nodev,relatime,size=489124k,nr_inodes=122281,mode=755)
tmpfs on /dev/shm type tmpfs (rw,nosuid,nodev,noexec,relatime,size=489124k,nr_inodes=122281)
tmpfs on /userdata/bluetooth type tmpfs (rw,nosuid,nodev,relatime,size=489124k,nr_inodes=122281,mode=755)
devtmpfs on /mnt/sdcard/Roms/.32bit_chroot/dev type devtmpfs (rw,relatime,size=477572k,nr_inodes=119393,mode=755)
tmpfs on /mnt/sdcard/Roms/.32bit_chroot/run type tmpfs (rw,nosuid,nodev,relatime,size=489124k,nr_inodes=122281,mode=755)
```
So, if you don't like /run, you can move it to /tmp or even to /dev/shm
Also, as I can see, there is only one normal partition, mounted as RW and it's /userdata
```
mount | grep rw | grep -E -v "sys|proc|tmpfs|mmcblk"
devpts on /dev/pts type devpts (rw,relatime,gid=5,mode=620,ptmxmode=666)
/dev/mtdblock4 on /userdata type ext4 (rw,relatime)
adb on /dev/usb-ffs/adb type functionfs (rw,relatime)
```
That means you can't accidetally coorrupt rootfs on flip

PR #459, opened 2025-04-27
Fix emufresh.py for PICO8 and improperly changing non configs

chrisj951:


ryanmsartor:
Tested this by removing my pico8 bins from my BIOS folder to verify the system disappeared; then added them back to verify it showed back up.

PR #468, opened 2025-04-29
Remove pausing for non A30 devices in RA

chrisj951:
It was causing weird lag / behavior of not actually unpausing


dhoizner (approved):
pause is 💀 long live pause

PR #475, opened 2025-05-01
fix devices not connecting

00svd00:
We also do not need to remove /tmp/bluetooth_ready
It's deleted automatically


ryanmsartor:
Thanks for the PR @00svd00 . Can you please give a little bit of detail as to what this fixes, and under what conditions you tested? I have been able to connect and use my bluetooth controllers without this (though it is notable that I have not tested BT audio without it yet).

00svd00:
>  I have been able to connect and use my bluetooth controllers 
Yeah. Controllers - no problem. But A2DP devices (bluetooth audio) will not work without it.
I've found theese commands while investigating keymon binary with HEX editor

ryanmsartor (approved):
Thanks for explaining! Looks good, let's try it out!

PR #511, opened 2025-05-11
Boxart Scraper: Fix height of exit image/text on flip

Dark98:
Fixed the Exit Image/Text being cut off on the flip

i don't own any other devices to check if their correct, I just set all planned supported devices to match the old setting


ryanmsartor:
Hey @Dark98 if you rename those variables to reflect that they control scaling rather than height I'll go ahead and merge this

Dark98:
> Hey @Dark98 if you rename those variables to reflect that they control scaling rather than height I'll go ahead and merge this. Actually, you can also just remove the platform check and make the change globally.

Done :)

PR #690, opened 2025-11-03
[WIP] Support second sdcard slot on Flip

### LINKED ISSUE #547: Doesn't see external SD card (Samsung EVO plus 512)
Doesn't detect the external SD card even through the file manager, while the stock firmware sees the games

Lawstorant:
Mounts appropriate folders from second SD card. Probably needs a watchdog to support hotplug

Fixes #547


Lawstorant:
It's a bug in game system detection

PR #691, opened 2025-11-03
Fix checking for roms in all viable folders for a given game system

### LINKED ISSUE #547: Doesn't see external SD card (Samsung EVO plus 512)

Lawstorant:
Fixes #547 

Game systems won't show up if there's an empty game system folder on the main SD card and that's the case for every system after spruceOS install. This fix restores proper behavior and now the system will properly show up in UI when roms are detected on the Second SD card or any other viable roms folder.


Lawstorant:
Moved to: https://github.com/chrisj951/MainPyUI/pull/44

PR #1078, opened 2026-01-11
feat: initial display write to kernel proc interface instead of keymon

xXJSONDeruloXx:


ryanmsartor:
@chrisj951 could the mismatch just be between the mapping / curves between internal and external values each implementation uses?

chrisj951:
> @chrisj951 could the mismatch just be between the mapping / curves between internal and external values each implementation uses?

No, the cd/m^2 value of white is changing in this version. On the stock one it doesn't. I also have never been able to get this version to keep the cd/m^2 brightness, while getting the colors i like

PR #1092, opened 2026-01-11
Add ZIP archive creation to nightly workflow

Sundownersport:
Added ZIP archive creation step alongside 7z archive in nightly workflow.

### DIFF
--- a/.github/workflows/create-nightly.yml
+++ b/.github/workflows/create-nightly.yml
@@ -1,11 +1,7 @@
 name: Create Nightly
 
 on:
-#  schedule:
-#    - cron: '10 10 * * *'
   workflow_dispatch:
-#  push:
-#    branches: [ Development ]
 
 jobs:
   create-7z:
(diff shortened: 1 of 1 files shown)


.github/workflows/create-nightly.yml
```diff
@@ -182,21 +168,12 @@ jobs:
...
         run: |
-          # Get all releases and filter out beta releases
           RELEASES=$(gh api repos/spruceUI/spruceOSNightlies/releases --paginate | \
             jq -r '.[] | select(.tag_name | contains("-Beta") | not) | .id')
-          
-          # Convert to array and keep only releases after the first 3
           COUNT=0
           for RELEASE_ID in $RELEASES; do
             COUNT=$((COUNT + 1))
             if [ $COUNT -gt 3 ]; then
```
Sundownersport, line 176: can we change this to `if [ $COUNT -gt 14 ]: then`

with as fast as we move it might be smart to keep a few more releases in case people need to roll back

.github/workflows/create-nightly.yml
```diff
@@ -1,11 +1,7 @@
 name: Create Nightly
 
 on:
-#  schedule:
-#    - cron: '10 10 * * *'
   workflow_dispatch:
-#  push:
-#    branches: [ Development ]
```
xXJSONDeruloXx, line 8: Would we want on chron if new commit?

PR #1177, opened 2026-01-25
pixel2: Enable network services and fix repairSD

Hairo:


spruce/scripts/networkservices.sh
```diff
@@ -30,10 +34,18 @@ connect_services() {
 	fi
 
 	# SSH check
-	if [ "$ssh_enabled" = "True" ] && ! pgrep "dropbearmulti" >/dev/null; then
-		# Flag exists but service is not running, so start it...
-		log_message "Network services: Dropbear detected not running, starting..."
-		start_dropbear_process
+	if [ "$PLATFORM" = "Pixel2" ]; then
```
chrisj951: We used to have platform checks littered throughout all the files which made adding new devices much more difficult. Can you refactor some of these to work through the device files? (i.e. not add platform checks inside the non device files, and emu for now as those were never updated)

spruce/scripts/network/sshdFunctions.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+
+start_sshd_process() {
+    log_message "Starting sshd..."
+    systemctl start sshd
+    flag_add "sshd"
```
chrisj951: What is the point of this flag? Nothing seems to check it and a flag doesn't seem right (i.e. between boots the flag would be there even if sshd isn't running)

Hairo: You're right, after checking the flag doesn't do anything, i was trying to make it as close as possible as dropbear (it has a useless flag too).

PR #1184, opened 2026-01-28
pixel2: portmaster support and cleanup device config

Hairo:


spruce/scripts/emu/lib/ports_functions.sh
```diff
@@ -35,34 +35,43 @@ is_retroarch_port() {
...
-	
-    set_port_mode
 
-    is_retroarch_port
+    # Setup variables
     PORTS_DIR=/mnt/SDCARD/Roms/PORTS
     export HOME="/mnt/SDCARD/Saves/flip/home"
     export LD_LIBRARY_PATH="$PORTS_LD_LIBRARY_PATH:$LD_LIBRARY_PATH"
-    export PATH="/mnt/SDCARD/spruce/flip/bin/:$PATH"
+
```
chrisj951, line 55: Mind adding a comment here that when PM is updated to detect spruce this should go away? And it's due to incorrectly detecting as rocknix for pixel2 here?

PR #1186, opened 2026-01-29
For SSH fix at some devices (MiyooMini etc)

ljhmd0825:


spruce/bin/SSH/dropbear-wrapper.sh
```diff
@@ -5,7 +5,7 @@ cd /mnt/SDCARD
 
 # TODO: this probably doesn't work outside the a30?
 if [ "$SSH_ORIGINAL_COMMAND" = "/usr/libexec/sftp-server" ]; then
-    if [ "$PLATFORM" = "A30" ]; then
+    if [ "$PLATFORM" = "A30" -o "$PLATFORM" = "MiyooMini" ]; then
```
chrisj951, line 8: For future updates this is better as 
"$ARCH" = "armhf"

PR #1246, opened 2026-02-08
Improve config reload and volume handling - TSPS

CatalyticArkun:
Add robust reload logic and volume sync/controls. SystemConfig.reload_config now retries on JSON/OSError, treats empty files as decode errors, logs warnings/errors, and uses a short retry delay. TrimUISmartProS: add math/re imports, reduce stock config watcher interval, convert between system (0-100) and UI (0-20) volume scales, implement _set_volume to call amixer (with a fallback), update system and stock JSON configs, notify Display of changes, and provide get_real_volume parsing. change_volume now adjusts UI steps and delegates to _set_volume. Misc: minor refactors and extra logging for error cases.

This needs to get tested on other systems and probably get refactored to assist Brick/TSP logic as well.

### DIFF
--- a/App/PyUI/main-ui/devices/trimui/trim_ui_smart_pro_s.py
+++ b/App/PyUI/main-ui/devices/trimui/trim_ui_smart_pro_s.py
@@ -1,6 +1,8 @@
 import subprocess
 import json
 import os
+import math
+import re
 from pathlib import Path
 import threading
 import time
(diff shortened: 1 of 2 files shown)


App/PyUI/main-ui/devices/trimui/trim_ui_smart_pro_s.py
```diff
@@ -138,29 +197,51 @@ def get_audio_system(self):
...
+                data = json.load(f)
+
+            old_volume = self.mainui_volume
+            self.mainui_volume = data.get("vol")
+            volume_changed = (old_volume != self.mainui_volume)
+            try:
+                system_volume = self._ui_to_system_volume(self.mainui_volume)
+                self.system_config.reload_config()
+                self.system_config.set_volume(system_volume)
+                self.system_config.save_config()
```
chrisj951, line 235: I think theres some confusion on what should be happening. Can you minimize this PR to just the needed changes + specify what they're doing?

e.g. if mainui config volume changes it implies we should never be touching the volume. So there's no need update our config.

CatalyticArkun:
There are 5 lights.

PR #1273, opened 2026-02-14
Turn off the screen when charging and idle

Hairo:
Configurable, works in menu and in game, other devices need to implement `turn_on_screen` and `turn_off_screen`. 

This commit also cleans up `idlemon_mm.sh`, it had unused parameters and `reapply` was always called so the parameters check was removed.

### DIFF
--- a/Saves/spruce/spruce-config.json
+++ b/Saves/spruce/spruce-config.json
@@ -92,6 +92,7 @@
                 "selected": "Off"
             },
             "idlemonChargingInMenu": {
+                "devices":["GKD_PIXEL2"],
                 "display": "Turn screen off when charging in menu",
                 "description": "only applies in main UI",
                 "options": [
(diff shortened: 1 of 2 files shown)


ryanmsartor:
Looks good, but I think I'd like to wait to merge until after the 4.1.0 stable release, so as to not have to mess with implementing those two functions in a rush

Saves/spruce/spruce-config.json
```diff
@@ -91,6 +91,29 @@
                 ],
                 "selected": "Off"
             },
+            "idlemonChargingInMenu": {
```
chrisj951, line 94: Can limit these to just pixel2 until we add support for others (if we even want to)

                "devices": ["GKD_PIXEL2"]

ryanmsartor: Thought about this. Guess it would be ok to not account for that in the idlemon code, as long as the defaults are set to not change the current behavior

Hairo: Alright, done.

ryanmsartor:
Perfect, thanks Hairo!

PR #1298, opened 2026-02-18
Replace all sed -i with jq or atomic sed

Sundownersport:
## Summary
- Replace all 47 `sed -i` instances across 23 files with crash-safe alternatives
- JSON files (system.json vol/backlight/wifi, config.json label toggling) now use `jq` for proper parsing instead of fragile regex
- Non-JSON files (config/INI, XML, text, CI) use atomic sed: `sed '...' "$f" > "$f.tmp" && mv "$f.tmp" "$f"`
- Consecutive `sed -i` calls on the same file are combined into single atomic operations

## Test plan
- [ ] Verify volume changes persist across reboot on each device (Flip, A30, Pixel2, SmartProS, TrimUI)
- [ ] Verify brightness changes persist across reboot on each device
- [ ] Verify OTA update checker correctly shows/hides app labels
- [ ] Verify EZ Updater and Firmware Update apps toggle visibility correctly
- [ ] Verify RetroArch config updates still apply (PerfectOverlays, per-game settings)
- [ ] Verify Syncthing config repair works on first launch
- [ ] Verify Gallery, ThemeGarden, BootLogo, GameNursery app launches work
- [ ] Verify PICO-8 screen transform applies correctly per platform
- [ ] Confirm `grep -r 'sed -i' --include='*.sh' --include='*.yml'` returns no results

### DIFF
--- a/.github/workflows/version-updater.yml
+++ b/.github/workflows/version-updater.yml
@@ -22,13 +22,13 @@ jobs:
           
       - name: Update autorun.inf
         run: |
-          sed 's/spruce v[0-9]\+\.[0-9]\+\.[0-9]\+/spruce v${{ steps.get-version.outputs.version }}/g' autorun.inf > autorun.inf.tmp && mv autorun.inf.tmp autorun.inf
-
+          sed -i 's/spruce v[0-9]\+\.[0-9]\+\.[0-9]\+/spruce v${{ steps.get-version.outputs.version }}/g' autorun.inf
+          
       - name: Update language files
         run: |
-          find miyoo/res/lang -type f -name "*.lang" -exec sh -c 'for f; do sed "s/spruce v[0-9]\+\.[0-9]\+\.[0-9]\+/spruce v${{ steps.get-version.outputs.version }}/g" "$f" > "$f.tmp" && mv "$f.tmp" "$f"; done' _ {} +
-          find miyoo355/app/lang -type f -name "*.lang" -exec sh -c 'for f; do sed "s/spruce v[0-9]\+\.[0-9]\+\.[0-9]\+/spruce v${{ steps.get-version.outputs.version }}/g" "$f" > "$f.tmp" && mv "$f.tmp" "$f"; done' _ {} +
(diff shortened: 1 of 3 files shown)


.github/workflows/version-updater.yml
ryanmsartor, line 1: This one can be left in its previous form. It never runs on device and is easier to read the way it was before.

spruce/scripts/network/syncthingFunctions.sh
ryanmsartor, line 1: We need to make sure that busybox sed on each of our devices actually supports the -e argument

App/-Updater/updater.sh
ryanmsartor, line 1: Double check what this jq call does. I'm not positive it's wrong because I'm not the best with jq without a reference, but it seems like this wouldn't do what we want

ryanmsartor: Actually it might be alright on second read.... but double check

spruce/scripts/platform/device_functions/SmartProS.sh
```diff
@@ -85,7 +85,7 @@ set_volume() {
 
         if [ "$current_volume" -ne "$new_vol" ]; then
             save_volume_to_config_file "$new_vol"
-            sed -i "s/\"vol\":[[:space:]]*[0-9]\+/\"vol\": $new_vol/" /mnt/UDISK/system.json
+            jq ".vol = $new_vol" /mnt/UDISK/system.json > /mnt/UDISK/system.json.tmp && mv /mnt/UDISK/system.json.tmp /mnt/UDISK/system.json
```
ryanmsartor: Revert this one to sed because jq will fail if stock breaks this copy of system.json, which it loves to randomly do

spruce/scripts/platform/device_functions/trimui_a133p.sh
```diff
@@ -65,7 +65,7 @@ set_volume() {
 
         if [ "$current_volume" -ne "$new_vol" ]; then
             save_volume_to_config_file "$new_vol"
-            sed -i "s/\"vol\":[[:space:]]*[0-9]\+/\"vol\": $new_vol/" /mnt/UDISK/system.json
+            jq ".vol = $new_vol" /mnt/UDISK/system.json > /mnt/UDISK/system.json.tmp && mv /mnt/UDISK/system.json.tmp /mnt/UDISK/system.json
```
ryanmsartor: Revert to sed for things touching stock's system.json

ryanmsartor (changes requested):
I think it's all good except the few things I specifically commented on, especially the stock system.json stuff

PR #1397, opened 2026-03-18
Add power-mode and observability hooks

CatalyticArkun:
Introduce a power mode runtime module and add observability/coordination hooks across scripts. Key changes:

- Add power_mode.sh: implements a locked, resumable power-mode state file and helpers for running/sleep/waking/shutdown_pending with validation and generation tracking.
- Refactor archiveUnpacker.sh: add state file, pre_cmd handoff (foreground/background), resumable failure states, robust silent-lock handling, UI visibility flags, more detailed logging and emit traces via system-emit; ensure cleanup and exit tracing.
- Update firstboot.sh: gate completion on unpack coordinator state, add UI hold flags, timeouts, and tracing; use shared show_firstboot_screen helper.
- Update helperFunctions.sh, low_power_warning.sh, networkservices.sh and various device platform scripts (AnbernicXXCommon, MiyooMini, Pixel2, SmartProS, trimui_a133p) to emit structured events (brightness/audio/network/power) and improve PyUI writer lifecycle reporting.
- Add new scripts: power_mode.sh, system-emit (executable) and trace.sh (for tracing/emit support).
- principal.sh: add autoresume protections and validation for /tmp/cmd_to_run.sh.

Overall this improves robustness for unpacking/firstboot workflows, adds observability and tracing, and provides a canonical power-mode contract for coordinated shutdown/sleep handling.


.github/workflows/version-updater.yml
ryanmsartor, line 1: are there any actual differences here? Is this just a line ending thing?

spruce/scripts/emu/lib/scummvm_functions.sh
ryanmsartor, line 1: are these changes inherited from Sun's branch?

.github/workflows/version-updater.yml
CatalyticArkun: No actual difference there - just line endings

spruce/scripts/emu/lib/scummvm_functions.sh
CatalyticArkun: I believe these are from overhauling first boot, but I'm not sure if first boot actually requires these.

.github/workflows/version-updater.yml
CatalyticArkun: Reverted the line ending noise

spruce/scripts/helperFunctions.sh
```diff
@@ -17,6 +17,7 @@
 export FLAGS_DIR="/mnt/SDCARD/spruce/flags"
 export MESSAGES_FILE="/var/log/messages"
 POWER_OFF_SCRIPT="/mnt/SDCARD/spruce/scripts/save_poweroff.sh"
+export SYSTEM_EMIT="${SYSTEM_EMIT:-/mnt/SDCARD/spruce/scripts/system-emit}"
```
ryanmsartor, line 19: can we keep this one definition of SYSTEM_EMIT here in helperFunctions and remove all the others? Pretty much every script in spruce sources helperFunctions.sh. I think the duplicate definitions of this end up just being noise.

spruce/scripts/runtime.sh
```diff
@@ -10,6 +10,43 @@
 
 . /mnt/SDCARD/spruce/scripts/helperFunctions.sh
 . /mnt/SDCARD/spruce/scripts/runtimeHelper.sh
+. /mnt/SDCARD/spruce/scripts/trace.sh
+SYSTEM_EMIT="${SYSTEM_EMIT:-/mnt/SDCARD/spruce/scripts/system-emit}"
+
+UNPACK_STATE_FILE="/mnt/SDCARD/Saves/spruce/unpacker_state"
+
+read_unpack_state() {
```
ryanmsartor: Can we pull these function definitions into runtimeHelper.sh to try to keep runtime.sh succinct? Readability is a priority

spruce/scripts/runtime.sh
```diff
@@ -35,22 +74,49 @@ unstage_archives_wanted
...
 fi
 
 /mnt/SDCARD/spruce/scripts/set_up_swap.sh &
 
 launch_startup_watchdogs
+"$SYSTEM_EMIT" process runtime "STARTUP_WATCHDOGS_LAUNCHED" "runtime.sh" "startup_watchdogs launched" || true
 
 # check whether to auto-resume into a game
+auto_resume_staged=0
 if flag_check "save_active"; then
```
ryanmsartor, line 113: It seems unnecessarily circuitous to check for a flag to determine whether to create a second flag, then delete that first flag, then check that second flag when actually executing the autoresume logic? I think you're on the right path with moving autoresume into principal, but I think the execution could be streamlined.

ryanmsartor (changes requested):

spruce/scripts/system-emit
```diff
@@ -0,0 +1,98 @@
+#!/bin/sh
+
```
ryanmsartor, line 2: I think right at the top of this script would be a good place to gate all of this new logging functionality behind a flag. Something like

`if ! flag_check "system-emit"; then exit 0; fi`

This way, in a production release we can minimize any potential performance impacts of this code, but during testing we can intentionally enable it to aid in debugging.

spruce/scripts/runtime.sh
CatalyticArkun: I agree with the concern. My intent with the split was to keep save_active as the persistent cross-boot trigger from shutdown, and use the staged flag only as the one-shot handoff so principal.sh remains the canonical launcher. That said, I think your point is right that the transition can be streamlined. If that approach sounds good to you, I can tighten the implementation in the next pass so the save_active -> staged handoff is more direct and the extra bookkeeping is reduced.

spruce/scripts/helperFunctions.sh
```diff
@@ -901,18 +918,43 @@ set_network_proxy() {
     fi
 }
 
+_log_scummvm_display_text() {
```
ryanmsartor: I'm still a bit confused as to why the scummvm stuff is in here. Truthfully I'm out of the loop with spruce's scummvm implementation currently, but I am a bit concerned about how these changes will play with SundownerSport's recent slew of scummvm-related changes. Also not sure about the use of the PortMaster icon here - didn't Sun just make some changes to obfuscate the scummvm extraction on firstboot? Would it be possible to remove these particular changes into a separate PR if they are indeed needed?

spruce/scripts/runtime.sh
ryanmsartor: Sounds good, please do :+1:

spruce/scripts/helperFunctions.sh
CatalyticArkun: From what I've seen, four extractions occur in the firstboot/archive unpacker flow already:
- system assets
- themes
- portmaster (on supported platforms)
- scummvm (on supported platforms)

I think the scummvm updates were incoming _while_ I was doing the firstboot refactor which is why it's confusing.  I recommend holding on this one until I reconcile the branch with the current dev branch state

#blame_sun

spruce/scripts/system-emit
CatalyticArkun: I agree with this change, and I think we should take it one step further: instead of using the normal flag_check path here, I’d prefer to gate system-emit off a tmp-backed flag so we’re not touching /mnt/SDCARD/spruce/flags on every trace emit. That would keep the production path cheap both in terms of logging overhead and SD access, while still giving us an explicit switch we can enable during debugging. If that sounds good, I can handle the next pass by adding the top-level gate in system-emit and making it tmp-only so the check stays in memory rather than pinging the SD card each time.

ryanmsartor: Sounds good, except then we potentially lose access to this script at boot time. To yes-and your recommendation, how about we use ONE flag_check() for a persistent flag early in runtime.sh, before the first call to system-emit,which will create the /tmp-backed flag if appropriate - and then within system-emit itself, check only that /tmp-backed flag.

CatalyticArkun: That sounds good to me. I agree that we still need one persistent check early in boot so trace availability can be decided before any system-emit calls happen, and then after that we should stay off the SD path for the hot emit case. I’m good with that shape for the next pass: do a single persistent flag_check in runtime.sh, mirror that decision into a /tmp-backed gate, and have system-emit consult only the tmp flag from that point forward.

👍

PR #1438, opened 2026-03-29
Fix WiFi not connecting on boot for TSP/Brick

Sundownersport:
## Summary
- TSP and Brick `startup_init()` only set display settings — WiFi was never proactively started, unlike the Flip which calls `start_wifi_services()` during init
- This meant WiFi relied on the shell-side background job plus the passive `monitor_wifi` 10-second poll loop, which could take up to ~60 seconds to react if something went wrong
- Now matches Flip behavior: if WiFi is enabled, stops stale services and starts fresh during `startup_init()`


chrisj951:
I dont think this fixes the real issue which needs to be fixed here:
https://github.com/spruceUI/spruceOS/blob/217cc5347d14c427b16b3326314afa145a2bce3a/spruce/scripts/runtime.sh#L27

This WILL probably fix wifi, but will make it take longer to startup than fixing it in the above

PR #1522, opened 2026-05-03
update libadvdrastic.so for tsps and tsp

trngaje:
update libadvdrastic.so for tsps and tsp


Sundownersport:
@trngaje can you target the Development branch instead please?

PR #1549, opened 2026-06-05
Sync Brick Fn actions and the volume keys with spruce state

Felixmil:
The TrimUI Brick "Fn Key and Switch Settings" actions are stock scripts that write to firmware sinks spruce does not use, so their effects did not stay in sync with spruce's own state. This reworks those actions to go through spruce's own helpers, so the hardware, the stored value, and the in-UI sliders all move together. It also stops the audio actions from triggering the stock volume popup, which was redundant with the switch's own on-screen toast, and adds a watcher so the in-UI volume bar follows the physical volume keys.

This builds on PR #1573 (the spruce-native Fn key watchdog) and should merge after it. This branch is stacked on top of that one, so until #1573 merges its commits also appear in this PR's diff; once #1573 lands, only the changes described below remain. The Fn key actions here only reach users once that watchdog is dispatching Fn presses; the volume key sync works on its own regardless.

## LED toggle action

The "LED ON/OFF toggle" action wrote `/tmp/system/enable_led` for the stock firmware daemons to act on. spruce drives the RGB itself through `/sys/class/led_anim` (see `enable_or_disable_rgb_trimui` / `set_rgb_in_menu`), so under spruce the stock script could switch the LEDs off but never restore them, because nothing re-applied the configured colour and effect. It now toggles the `led_anim` enable gate directly and, when turning back on, re-applies the user's configured colour and effect via spruce's `set_rgb_in_menu`, so on and off both work and "on" matches what spruce would show.

## Backlight action

The "LCD brightness switcher" action wrote the new level to `/tmp/system/set_brightness` for the stock `hardwareservice` to apply, which spruce does not consume, so the brightness slider and stored value did not follow the key. It now sources `helperFunctions.sh` and cycles the backlight with spruce's `current_backlight` / `set_backlight`, stepping through 1, 4, 7, 10 and wrapping back to 1. Because `set_backlight` records the value spruce reads elsewhere, the UI slider updates in step.

## Silent Mode action

Silent Mode hard-muted the speaker via `/sys/class/speaker/mute` but never told spruce, so the in-UI volume value stayed at its old level while muted. It now also drops the volume to 0 through spruce's `set_volume` on enter (so the volume value reflects the mute) and restores the prior volume on exit. The speaker hard-mute is kept as a backstop.

## Quiet Mode action

Quiet Mode poked an ALSA mixer control directly (`tinymix set 9 N`), which spruce neither reads nor respects, so nothing audible changed and the volume slider stayed wrong. It now lowers the volume to a low but audible level through spruce's `set_volume` on enter, saving the current volume first, and restores that saved volume on exit, mirroring the Silent Mode approach.

## Volume popup suppression

In game (when MainUI is not running) `set_volume` shows the stock volume OSD popup. When Silent or Quiet was toggled by the switch this popped up alongside the switch's own toast, which was redundant. `set_volume` gains an optional third argument that controls the popup (defaulting to shown, so every existing caller is unchanged), and the Silent and Quiet actions opt out of it. The physical volume keys still show their popup as before.

## Volume key sync

The physical Volume +/- keys never moved spruce's stored volume (`.vol`) in step with the audio, for a different reason on each TrimUI device. On the Brick they are handled entirely by the stock firmware, which adjusts the audio and writes the new level to `/tmp/system/set_volume` on every key event (autorepeat included); `Brick.cfg` sets no `EVENT_PATH_VOLUME`, so `buttons_watchdog.sh` never sees those keys, and spruce never read that file back. On the Smart Pro and Smart Pro S spruce does own the keys, but its hold loop steps with `SAVE_TO_CONFIG=false` and writes `.vol` only when the key is released. Either way the in-UI volume bar, which reads `.vol`, sat still while a key was held. A small watchdog (`spruce/scripts/volume_sync_watchdog.sh`) mirrors writes of `/tmp/system/set_volume` straight into `.vol`, a single in-place edit rather than a call back into `set_volume` (which would re-poke the command file and add a redundant flash write), so the bar tracks the hardware keys including while held. PyUI's config watcher polls at 50ms on all three devices so it can follow each step.

The Smart Pro and Smart Pro S also mute the speaker amp for the first seconds of boot to avoid a startup pop, and simulate volume keypresses inside that window. The watchdog's Silent Mode unmute is therefore gated on Silent Mode's own `/tmp/system/muted` marker, so it only ever clears the mute Silent Mode set and leaves the boot mute alone.

## Watchdog launch

The volume sync watchdog applies to every TrimUI device, so it is launched and pinned from `launch_trimui_startup_watchdogs` in `trimui_delegate.sh`, which the Brick, Smart Pro and Smart Pro S all source. #1573's Fn key watchdog is Brick-only (the Fn keys report as `B_L3` / `B_R3`, which are the stick clicks on the Smart Pro), so `Brick.sh` overrides `launch_startup_watchdogs` to call the shared one and add just that.

## Validation

Validated on a physical Brick (FW 1.1.1): the Fn LED toggle turns the RGB off and back on (restoring the configured colour), the backlight cycle runs 1 to 4 to 7 to 10 to 1 with the stored value updating each step, Silent Mode drops the volume value to 0 on enter and restores the previous level on exit, Quiet Mode drops the volume to a low level on enter and restores it on exit, in game the switch no longer shows the stock volume popup for Silent or Quiet while the volume keys still do, and holding the Volume +/- keys ramps the in-UI volume bar smoothly in sync with the audio. Both watchdogs run after boot with no keymon present.

The volume sync watchdog is untested on the Smart Pro and Smart Pro S: the logic is device-agnostic and reviewed against their scripts, but I have no such unit to check the bar ramp or the startup pop on, so a second pair of eyes there would be welcome.

### DIFF
--- a/App/PyUI/main-ui/devices/trimui/trim_ui_brick.py
+++ b/App/PyUI/main-ui/devices/trimui/trim_ui_brick.py
@@ -51,7 +51,7 @@ class TrimUIBrick(TrimUIDevice):
             # it. A bare stat() per tick is cheap; the granularity-repeat window
             # stays ~2s since it counts down in seconds.
             self.config_watcher_thread, self.config_watcher_thread_stop_event = FileWatcher().start_file_watcher(
-                "/mnt/SDCARD/Saves/trim-ui-brick-system.json", self.on_system_config_changed, interval=0.05, repeat_trigger_for_mtime_granularity_issues=True)
+                "/mnt/SDCARD/Saves/trim-ui-brick-system.json", self.on_system_config_changed, interval=0.2, repeat_trigger_for_mtime_granularity_issues=True)
             if(PyUiConfig.enable_button_watchers()):
                 from controller.controller import Controller
                 #/dev/miyooio if we want to get rid of miyoo_inputd
--- a/App/PyUI/main-ui/devices/trimui/trim_ui_brick.py
+++ b/App/PyUI/main-ui/devices/trimui/trim_ui_brick.py
@@ -44,12 +44,6 @@ class TrimUIBrick(TrimUIDevice):
             self.ensure_wpa_supplicant_conf()
             threading.Thread(target=self.monitor_wifi, daemon=True).start()
             threading.Thread(target=self.startup_init, daemon=True).start()
-            # Poll briskly (50ms) so the volume bar tracks the hardware Volume
-            # keys closely: the firmware updates set_volume on every key event and
-            # the volume sync watchdog mirrors it into this file immediately, so a
-            # held key should see the bar chase the firmware OSD rather than trail
-            # it. A bare stat() per tick is cheap; the granularity-repeat window
-            # stays ~2s since it counts down in seconds.
             self.config_watcher_thread, self.config_watcher_thread_stop_event = FileWatcher().start_file_watcher(
                 "/mnt/SDCARD/Saves/trim-ui-brick-system.json", self.on_system_config_changed, interval=0.2, repeat_trigger_for_mtime_granularity_issues=True)
             if(PyUiConfig.enable_button_watchers()):
--- a/App/PyUI/main-ui/devices/trimui/trim_ui_brick.py
+++ b/App/PyUI/main-ui/devices/trimui/trim_ui_brick.py
@@ -44,8 +44,14 @@ class TrimUIBrick(TrimUIDevice):
             self.ensure_wpa_supplicant_conf()
             threading.Thread(target=self.monitor_wifi, daemon=True).start()
             threading.Thread(target=self.startup_init, daemon=True).start()
+            # Poll briskly (50ms) so the volume bar tracks the hardware Volume
+            # keys closely: the firmware updates set_volume on every key event and
+            # the volume sync watchdog mirrors it into this file immediately, so a
+            # held key should see the bar chase the firmware OSD rather than trail
+            # it. A bare stat() per tick is cheap; the granularity-repeat window
+            # stays ~2s since it counts down in seconds.
             self.config_watcher_thread, self.config_watcher_thread_stop_event = FileWatcher().start_file_watcher(
-                "/mnt/SDCARD/Saves/trim-ui-brick-system.json", self.on_system_config_changed, interval=0.2, repeat_trigger_for_mtime_granularity_issues=True)
+                "/mnt/SDCARD/Saves/trim-ui-brick-system.json", self.on_system_config_changed, interval=0.05, repeat_trigger_for_mtime_granularity_issues=True)
             if(PyUiConfig.enable_button_watchers()):
                 from controller.controller import Controller
                 #/dev/miyooio if we want to get rid of miyoo_inputd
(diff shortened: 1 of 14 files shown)


App/PyUI/main-ui/devices/trimui/trim_ui_brick.py
```diff
@@ -44,8 +44,14 @@ def __init__(self, device_name, main_ui_mode):
...
             threading.Thread(target=self.startup_init, daemon=True).start()
+            # Poll briskly (50ms) so the volume bar tracks the hardware Volume
+            # keys closely: the firmware updates set_volume on every key event and
+            # the volume sync watchdog mirrors it into this file immediately, so a
+            # held key should see the bar chase the firmware OSD rather than trail
+            # it. A bare stat() per tick is cheap; the granularity-repeat window
+            # stays ~2s since it counts down in seconds.
             self.config_watcher_thread, self.config_watcher_thread_stop_event = FileWatcher().start_file_watcher(
-                "/mnt/SDCARD/Saves/trim-ui-brick-system.json", self.on_system_config_changed, interval=0.2, repeat_trigger_for_mtime_granularity_issues=True)
+                "/mnt/SDCARD/Saves/trim-ui-brick-system.json", self.on_system_config_changed, interval=0.05, repeat_trigger_for_mtime_granularity_issues=True)
```
chrisj951: Can you leave this at 0.2 and fix the merge conflict please?  Otherwise lgtm

chrisj951:
> Both this branch's volume sync watchdog and https://github.com/spruceUI/spruceOS/pull/1573's Fn key watchdog are Brick-only, so they are launched and pinned from a single launch_startup_watchdogs in Brick.sh rather than the shared A133P base (trimui_a133p.sh), keeping device-specific code in the device file.

Do you know if this shouldn't also apply to the smartpro?

Felixmil:
Good call, I've widened it: the volume sync watchdog now runs on all three TrimUI devices.

The Fn watchdog stays Brick-only. The editor is gated to the Brick (`App/fn_editor/config.json`: `"devices": ["TRIMUI_BRICK"]`), and the Fn keys report as `1 317` / `1 318`, which on the Smart Pro are the L3/R3 stick clicks.

The volume one had a real gap everywhere, just for different reasons. The Brick sets no `EVENT_PATH_VOLUME`, so `buttons_watchdog.sh` never reads the volume keys and the firmware owns them, writing only `/tmp/system/set_volume`. The Smart Pro and Smart Pro S do read the keys, but their hold loop steps with `SAVE_TO_CONFIG=false` and writes `.vol` only on release, so the bar jumped instead of ramping. Mirroring `/tmp/system/set_volume` into `.vol` fixes all three.

So: the watchdog moved from `spruce/brick/` to `spruce/scripts/`, the launch moved into `launch_trimui_startup_watchdogs` in `trimui_delegate.sh` (sourced by all three, with `Brick.sh` overriding only to add the Fn watchdog), and the Smart Pro / Smart Pro S PyUI config polls drop to 50ms like the Brick's.

One thing the widening turned up: the Smart Pro and Smart Pro S mute the speaker amp for the first seconds of boot against the startup pop, and simulate volume keypresses inside that window. That would have tripped the watchdog's Silent Mode unmute and cut the pop guard short, so the unmute is now gated on Silent Mode's own `/tmp/system/muted` marker. The Brick never hit it because its `sendevent` gets an empty `EVENT_PATH_VOLUME`.

Not tested on hardware with this latest change

PR #1550, opened 2026-06-05
Name the Brick Fn switch popup after the configured action

Felixmil:
On the TrimUI Brick, flipping the physical Fn switch shows an on-screen popup that always reads a generic ON / OFF, giving no indication of what the switch just toggled. This makes the popup label the configured switch action instead (for example Silent: ON, Quiet: OFF, LED: OFF), so the feedback matches what the switch actually does. The stock firmware popups live in the read-only `/usr/trimui` partition, so spruce overlays its own versions at boot using the same bind-mount mechanism it already uses for other Brick firmware files.

## Popup scripts

Two replacement popup scripts live under `spruce/brick/fn_dip/`. Each derives the label from the active switch action, which is the single script present in `/usr/trimui/scene/`, by mapping that script's basename to a short name (LED, Quiet, Silent, Joystick). Anything unmapped falls back to the original ON / OFF, so unknown or future actions degrade gracefully. The lookup is self-contained POSIX shell with no dependency on jq, since the daemon runs these on every flip.

The LED action is the stock "LED off" toggle, where switch-on turns the LEDs off. Its label is therefore inverted relative to the switch state (switch-on shows `LED: OFF`, switch-off shows `LED: ON`) so the text reflects the LED state rather than the switch position. The other actions follow the switch state directly.

## Boot integration

`runtime_mounts_a133p` in the A133P device init now bind-mounts both popup scripts over their firmware copies in `/usr/trimui/apps/fn_editor/`, alongside the existing `/etc` and SDL bind-mounts and before `trimui_scened` (the daemon that invokes the popups) starts. This keeps the firmware partition untouched and leaves no trace when the SD card is removed.

This builds on the Fn switch dispatch path; the popup only appears while the switch events are being handled (see #1548).


spruce/scripts/platform/device_functions/trimui_a133p.sh
```diff
@@ -154,6 +154,12 @@ runtime_mounts_a133p() {
     mount -o bind "${SPRUCE_ETC_DIR}/group" /etc/group &
     mount -o bind "${SPRUCE_ETC_DIR}/passwd" /etc/passwd &
     /mnt/SDCARD/spruce/brick/sdl2/bind.sh &
+    # Show the configured switch action's name in the Fn switch popup instead
+    # of a generic ON/OFF. trimui_scened calls these popup scripts on every
+    # flip, so overlay our versions before the blobs start below.
+    FN_DIP_DIR="/usr/trimui/apps/fn_editor"
+    mount --bind /mnt/SDCARD/spruce/brick/fn_dip/show_fn_dip_on_msg.sh "${FN_DIP_DIR}/show_fn_dip_on_msg.sh" &
```
chrisj951: If this is only for brick can you move these into Brick.sh the device_init function underneath device_init_a133p please

Felixmil: thank you for your review, done in [ddce96e](https://github.com/spruceUI/spruceOS/pull/1550/commits/ddce96e8490a0fa0028b1b604038b6032a7f95ce)

spruce/brick/fn_dip/show_fn_dip_on_msg.sh
```diff
@@ -0,0 +1,40 @@
...
+STATE="ON"
+SCENE_SH=$(ls /usr/trimui/scene/*.sh 2>/dev/null | head -n1)
+BASE=$(basename "$SCENE_SH" 2>/dev/null)
+case "$BASE" in
+    # The LED action is "LED off": switch ON turns the LEDs OFF, so the popup
+    # text is inverted relative to the switch state.
+    com.trimui.ledc.sh)                 MSG="LED: OFF" ;;
+    com.trimui.quiet.sh)                MSG="Quiet: $STATE" ;;
+    com.trimui.silent.sh)               MSG="Silent: $STATE" ;;
+    com.trimui.joystick.sh|com.trimui.toggle.dpad_joystick.sh) MSG="Joystick: $STATE" ;;
```
chrisj951: At least this one looks to be for stock. Spruce has

/usr/trimui/scene/spruce_toggle_joystick.sh

spruce/scripts/platform/device_functions/Brick.sh
```diff
@@ -30,6 +30,20 @@ init_gpio_a133p() {
...
+    # flip, so overlay our Brick-only versions with a bind mount.
+    FN_DIP_DIR="/usr/trimui/apps/fn_editor"
+    mount --bind /mnt/SDCARD/spruce/brick/fn_dip/show_fn_dip_on_msg.sh "${FN_DIP_DIR}/show_fn_dip_on_msg.sh" &
+    fn_dip_on_pid=$!
+    mount --bind /mnt/SDCARD/spruce/brick/fn_dip/show_fn_dip_off_msg.sh "${FN_DIP_DIR}/show_fn_dip_off_msg.sh" &
+    fn_dip_off_pid=$!
+    # Wait only for the two bind mounts, not a bare `wait`: device_init_a133p
+    # leaves fire-and-forget background jobs running (the startup-volume helper
+    # sleeps then blocks on sendevent), and a bare `wait` would hang boot on them.
+    wait "$fn_dip_on_pid" "$fn_dip_off_pid"
```
chrisj951: I don't think this wait is needed at all

spruce/brick/fn_dip/show_fn_dip_off_msg.sh
```diff
@@ -0,0 +1,40 @@
...
+STATE="OFF"
+SCENE_SH=$(ls /usr/trimui/scene/*.sh 2>/dev/null | head -n1)
+BASE=$(basename "$SCENE_SH" 2>/dev/null)
+case "$BASE" in
+    # The LED action is "LED off": switch OFF turns the LEDs back ON, so the
+    # popup text is inverted relative to the switch state.
+    com.trimui.ledc.sh)                 MSG="LED: ON" ;;
+    com.trimui.quiet.sh)                MSG="Quiet: $STATE" ;;
+    com.trimui.silent.sh)               MSG="Silent: $STATE" ;;
+    com.trimui.joystick.sh|com.trimui.toggle.dpad_joystick.sh) MSG="Joystick: $STATE" ;;
```
chrisj951: At least this one looks to be for stock. Spruce has

/usr/trimui/scene/spruce_toggle_joystick.sh

chrisj951 (changes requested):
See comments

chrisj951:
Tested this change and it also works on brick pro with needed fixes

PR #1565, opened 2026-07-02
feat: add TrimUI PyUI media and fn settings

vigstudio:
## Why

TrimUI Smart Pro/S users need PyUI-native access to stock-like video playback and Fn/DIP switch settings without falling back to the stock TrimUI UI. The stock CPU scene scripts also fail under spruceOS because CPU sysfs controls are locked and some stock frequencies are invalid on Smart Pro S, so Fn slider CPU performance mode does not apply reliably.

## What Changed

- Added PyUI Video Player browse, search, and controls screens backed by the existing MEDIA launcher.
- Added PyUI Fn Key Settings for stock TrimUI DIP switch scene actions and joystick mode.
- Added English and Vietnamese strings for the new PyUI screens.
- Added Video Player and Fn editor icons, and made built-in app icons prefer app-folder icons.
- Updated backlight Fn script to write the device backlight directly and persist brightness in `/mnt/UDISK/system.json`.
- Added Smart Pro S CPU performance/save scene scripts that unlock and re-lock CPU sysfs controls and use valid CPU4 frequencies.
- Updated `buttons_watchdog.sh` to poll GPIO363 for Fn slider/DIP changes, run scene scripts, and stop prior scene script instances before starting new ones.

## Verification

- Ran `git diff --check` successfully.
- Ran `python -m py_compile` for `trimui_fn_settings_app.py`, `video_player_app.py`, and `app_menu.py` successfully.
- On TrimUI Smart Pro S over SSH, toggled the physical Fn slider and verified `/tmp/buttons_watchdog_dip.log` recorded `state 1` and `state 0` transitions.
- Verified CPU4 changes to `performance 1992000 1992000` for performance state and returns to `ondemand 1008000 1992000` for normal state.


spruce/scripts/buttons_watchdog.sh
```diff
@@ -154,6 +211,12 @@ getevent $EVENTS | while read line; do
                 brightness_up
             fi
         ;;
+        *"key $B_L3 1"*) # Left Fn key down
```
chrisj951: Does the fn-key have the same keycode as L3/R3? 

Even if it does this should be slightly refactored to

1) Have a new export for fn-key instead of reusing l3/r3
2) Make a call to fn_key_down/fn_key_up
3) in device.sh just make empty fn_key_down/up methods like all the others in there
4) move the run_fn_config  into SmartProS.sh and this call into fn_key_down/up methods from (3)

App/PyUI/main-ui/menus/app/app_menu.py
```diff
@@ -1,151 +1,218 @@
...
-                        self.show_all_apps = AppMenuPopup(self.show_all_apps).run_app_menu_popup(None)
-                elif(Theme.skip_main_menu() and ControllerInput.L1 == selected.get_input()):
-                    self.save_app_selection(selected)
-                    return ControllerInput.L1
-                elif(Theme.skip_main_menu() and ControllerInput.R1 == selected.get_input()):
-                    self.save_app_selection(selected)
-                    return ControllerInput.R1
-                        
-                    
+
```
chrisj951: Are most of the line changes just line ending?

App/PyUI/main-ui/menus/apps/video_player_app.py
```diff
@@ -0,0 +1,295 @@
...
+from views.grid_or_list_entry import GridOrListEntry
+from views.selection import Selection
+from views.view_creator import ViewCreator
+from views.view_type import ViewType
+
+VIDEO_EXTENSIONS = {
+    ".mp4", ".mkv", ".avi", ".mov", ".flv", ".ts", ".webm",
+    ".m4v", ".wmv", ".mpeg", ".mpg", ".3gp",
+}
+VIDEO_ROOT = "/mnt/SDCARD/Roms/MEDIA"
```
chrisj951: Please update these to go through py-ui-config.json (located in App/PyUI/py-ui-config.json) and loaded through PyUiConfig.py

vigstudio:
Updated branch `vigstudio:feature/trimui-stock-settings`.

Changes addressed:
- Merged current `upstream/Development` and resolved conflicts in `English.json`, `Vietnamese.json`, and `app_menu.py`.
- Reduced `app_menu.py` conflict to functional app additions while keeping upstream i18n descriptions.
- Moved video root path into `App/PyUI/py-ui-config.json` and `PyUiConfig.get_video_root_path()`.
- Added device-level `fn_key_down` / `fn_key_up` hooks, SmartProS-specific Fn handling, and separate `B_FN_LEFT` / `B_FN_RIGHT` exports instead of using `B_L3` / `B_R3` directly in `buttons_watchdog.sh`.

Validation run:
- `python -m json.tool App/PyUI/lang/English.json`
- `python -m json.tool App/PyUI/lang/Vietnamese.json`
- `python -m json.tool App/PyUI/py-ui-config.json`
- `python -m py_compile App/PyUI/main-ui/menus/app/app_menu.py App/PyUI/main-ui/menus/apps/video_player_app.py App/PyUI/main-ui/menus/apps/trimui_fn_settings_app.py App/PyUI/main-ui/utils/py_ui_config.py`
- `bash -n spruce/scripts/buttons_watchdog.sh spruce/scripts/platform/device.sh spruce/scripts/platform/device_functions/SmartProS.sh App/fn_editor/com.trimui.cpuperformance.sh App/fn_editor/com.trimui.cpusave.sh App/fn_editor/com.trimui.switch.backlight.sh`

Note: `git diff --check` still reports trailing whitespace from the `upstream/Development` i18n merge content; I left it untouched to avoid a broad whitespace-only cleanup in this PR.

vigstudio:
Created duplicate PyUI-only PR for MainPyUI:
https://github.com/chrisj951/MainPyUI/pull/91

vigstudio:
Updated `vigstudio:feature/trimui-stock-settings` against current `upstream/Development`.

## Merge
- Merged latest `upstream/Development` (includes screensaver, SmartProS button settings, Brick/BrickPro fn work).
- Resolved conflicts in:
  - `App/PyUI/py-ui-config.json` — kept both `videoRootPath` and `screensaverTimeoutSec`
  - `App/PyUI/main-ui/utils/py_ui_config.py` — kept `get_video_root_path()`
  - `App/fn_editor/com.trimui.switch.backlight.sh` — adopted upstream spruce-native `set_backlight` path
  - `App/fn_editor/config.json` — kept Brick + BrickPro device list from upstream

## Review feedback status (still in place)
1. Fn keys use `B_FN_LEFT` / `B_FN_RIGHT` + `fn_key_down` / `fn_key_up` hooks (SmartProS-specific).
2. Video root goes through `py-ui-config.json` / `PyUiConfig.get_video_root_path()`.
3. `app_menu.py` remains a focused functional diff for the new apps.

## Validation
- `python -m json.tool App/PyUI/py-ui-config.json`
- `python -m py_compile` for `py_ui_config.py`, `app_menu.py`, `video_player_app.py`, `trimui_fn_settings_app.py`
- `bash -n` for `buttons_watchdog.sh`, `device.sh`, `SmartProS.sh`, `com.trimui.switch.backlight.sh`

Sundownersport (commented):
Thanks for this, and sorry it's taken a while to get eyes on it. The write-up is genuinely one of the clearest PR descriptions we've had — having the actual on-device verification steps in there made reviewing it much easier. The problem is real too: Fn slider CPU mode not applying on Smart Pro S has come up before, and PyUI-native media/Fn screens are something people have asked for.

I'd like to get this in. Here's what I think needs to happen first.

## The one that blocks merge

`buttons_watchdog.sh` is shared by every device, and `dip_switch_bg &` is started unconditionally at top level. On anything that isn't a Smart Pro S — A30, Mini, Flip, Anbernic — that leaves a 5 Hz poll loop running forever with nothing to read. The part I'd want to avoid shipping is this:

```sh
echo 363 > /sys/class/gpio/export
echo -n in > /sys/class/gpio/gpio363/direction
```

GPIO 363 is a Smart Pro S pin. On another SoC that number is something else entirely, and we'd be exporting and reconfiguring it on every boot. Could this go behind a device function, the way the rest of the platform-specific behaviour works? Starting the poller from `device_functions/SmartProS.sh` would keep it off hardware that has no business running it.

## Fn key dispatch — we have a pattern for this now

This is entirely on our timing rather than yours: `spruce/brick/fnkey_watchdog.sh` landed on 2026-07-07, five days after you opened this PR, so it wasn't there when you started. It solves the same problem — spruce doesn't run keymon, so the Fn keys do nothing — as a standalone watchdog that reads `/usr/trimui/fnkeys/f1key_launch`, has an outer restart loop, uses `getevent -pid $$`, and logs through `log_message`.

Right now this PR would give two sibling TrimUI devices two different implementations with two different data models (`f1key_launch` plain text vs `f1key.json` + `jq`). Would you be up for converging on that one? Smart Pro S mostly needs the launcher wired up plus your DIP handling, which the Brick doesn't have.

## Thermal profile writes

I don't think these are landing the way you'd expect:

- `set_thermal_profile sport` / `conservative` — the profiles directory only has `high.json`, `low.json`, `medium.json`, `off.json`, so those names don't resolve to anything.
- `active_profile` is normally written by `spruce/smartpros/bin/update-thermal-watchdog-to-setting`, which stops the watchdog binary, writes the file, then restarts it. Writing the file on its own means the running watchdog never re-reads it.
- It also puts the file out of sync with `spruce-config.json`, so the Thermal Control option in System Settings would disagree with what's on disk and quietly overwrite it on the next settings change or reboot.

My suggestion would be to drop the thermal bits from this PR entirely and do them separately through that entry point — it keeps this PR much easier to reason about.

## Smaller things

- **`pkill -f "$(basename "$scene")"`** matches on full command line, so a generically-named scene script could take unrelated processes with it.
- **Debug logging** — `/tmp/buttons_watchdog_dip.log` and `_fn.log` grow unbounded in RAM, and `echo "$line" > /tmp/buttons_watchdog_last_event` is a file write on every input event on every device. I'm guessing these were for the SSH testing and just need pulling back out.
- **The lang files** got reindented from 4-space to 2-space, which turned ~40 real new strings into ~860 lines of diff (431 of the 432 removed lines are byte-identical to added ones). Same whitespace churn in `app_menu.py` and `py-ui-config.json`. Could you restore the original indentation and keep just the new keys? It'll make this much easier to review and stop it conflicting with other translation work.
- **Device gating** — `app_menu.py` hardcodes `("TRIMUI_SMART_PRO_S", "TRIMUI_SMART_PRO", "TRIMUI_BRICK")` in Python, but there's already the declarative route that your own `App/VideoPlayer/config.json` uses (`"devices": [...]`, read at `app_menu.py:91`). Worth using that for consistency. Also `TRIMUI_BRICK_PRO` is missing from the video player list — it's a supported device and would silently miss out.
- **`App/VideoPlayer`** ships with `"hide": true` and "Deprecated: use PyUI Video Player" from day one. If it's only there for the icon and the stock-player fallback, maybe worth saying so in a comment, or moving the icon somewhere less surprising.
- **`com.trimui.cpuperformance.sh` and `com.trimui.cpusave.sh`** are ~90% the same file. One script taking a mode argument would be easier to keep in step.

## One to double-check

The scripts set cpu4 directly to `performance 1992000 1992000`. Have you checked how that interacts with `set_smart` / `enforceSmartCPU.sh`? If spruce's CPU management reasserts afterwards, we'd get intermittent "my clocks reset themselves" reports that are miserable to track down. You may well have already confirmed this on device — if so, ignore me.

## Structure I liked

The `fn_key_down` / `fn_key_up` no-op in `device.sh` with per-device overrides is exactly right and matches how the rest of the platform layer works. And shipping Vietnamese alongside English is appreciated — not many contributors do that.

---

Heads up that Development has moved about 52 commits since your base, so this'll want a rebase whenever you pick it back up. No rush from my side, and happy to talk through any of the above if you'd rather approach it differently — particularly the Fn dispatch question, since that's more about which direction we standardise on than about anything being wrong with your version.

Sundownersport:
this is Claudes review btw, feel free to reach out with any questions or concerns!

chrisj951:
Closing due to inactivity

PR #1573, opened 2026-07-07
Dispatch Brick Fn keys with a spruce-native watchdog

### LINKED ISSUE #1440: Function Keys and slider not working in TrimUI Brick

Felixmil:
On the TrimUI Brick, the "Fn Key and Switch Settings" app lets you map an action to each of the two physical Fn keys, but pressing them did nothing. This adds Fn key support the spruce-native way: a small watchdog reads the Fn key presses and runs the mapped action, with no stock `keymon`.

This supersedes the earlier approach in #1548, which enabled Fn keys by adding the stock `keymon` daemon to the boot blob list. That PR was closed because spruce deliberately does not run `keymon` (it was removed across the project and replaced with spruce's own `getevent` watchdogs; `keymon` is kept only for the Miyoo Mini, where it is a different, stock component). This PR delivers the same feature using the existing watchdog pattern instead, so it fits that decision.

## Fn key watchdog

`spruce/brick/fnkey_watchdog.sh` reads Fn key-down events from `$EVENT_PATH_READ_INPUTS_SPRUCE` with `getevent` (the same idiom as `homebutton_watchdog.sh`) and runs the action the `fneditor` app mapped to each key. The chosen action is read fresh on every press from `/usr/trimui/fnkeys/f1key_launch` (left) and `f2key_launch` (right), so remapping in the editor takes effect immediately, and an unmapped key is ignored. The action scripts are stateless togglers invoked with no argument, so the watchdog just executes the launch path, exactly as `keymon` did.

On the Brick the left/right Fn keys report as `$B_L3` (`1 317`) and `$B_R3` (`1 318`) on `event3`; the Brick has no analog sticks, so those codes are exclusively the Fn keys. The node is read non-exclusively, alongside the home and buttons watchdogs that already read it, so there is no double-handling.

## Boot wiring

`launch_startup_watchdogs` in `trimui_a133p.sh` launches and CPU-pins the watchdog next to the common startup watchdogs, matching how the other durable watchdogs are started. The boot blob list is unchanged (no `keymon`).

## Validation

Tested on a physical Brick (FW 1.1.1) from a clean reboot: `keymon` is not running, `fnkey_watchdog.sh` starts and is pinned, and it coexists with the home, buttons, power, and volume watchdogs on the shared `event3` node. The right Fn key (mapped to the LCD brightness cycle) works end to end through spruce's own backlight control. The left Fn key's LED toggle dispatches correctly, but the stock LED action script is a no-op under spruce (it writes `/tmp/system/enable_led`, which the stock firmware daemons consume and spruce does not); making the physical RGB respond is a separate, pre-existing gap in that action script and is left for a follow-up.

Closes #1440

### DIFF
--- a/spruce/brick/fnkey_watchdog.sh
+++ b/spruce/brick/fnkey_watchdog.sh
@@ -1,0 +1,63 @@
+#!/bin/sh
+# Dispatch the Brick's two Fn keys by running the action the user picked in the
+# "Fn Key and Switch Settings" editor.
+#
+# On stock firmware the TrimUI keymon daemon reads the editor's per-key mapping
+# and runs the chosen action on each Fn press. spruce deliberately does not run
+# keymon (it replaced keymon everywhere with its own getevent watchdogs), so the
+# Fn keys did nothing. This watchdog is the spruce-native replacement for that
+# one job: watch the input node, and on an Fn key-down run the mapped action.
+#
+# Data model (all owned by the editor, /usr/trimui/apps/fn_editor/fneditor):
+#   /usr/trimui/fnkeys/f1key_launch  - one line, absolute path to the left  Fn action
+#   /usr/trimui/fnkeys/f2key_launch  - one line, absolute path to the right Fn action
+# The launch files are re-read on every press, so remapping in the editor takes
+# effect immediately with no restart. An unmapped key has no launch file and is
+# simply ignored.
+#
+# The action scripts are stateless togglers invoked with no argument (e.g.
+# com.trimui.toggle.ledc.sh flips the LED, com.trimui.switch.backlight.sh cycles
+# brightness), so this watchdog just executes the launch path as-is, exactly as
+# keymon did.
+#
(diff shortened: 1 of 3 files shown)


Felixmil:
Heads-up on ordering: PR #1549 (Brick Fn action + volume-key sync) is stacked on top of this one and should merge after it. That PR reuses the Fn key watchdog added here to dispatch its LED/backlight Fn actions, and moves its Brick-only watchdog launch into the same `launch_startup_watchdogs` in `Brick.sh` that this PR introduces.

chrisj951 (commented):
From what I can tell the switch toggle works without keymon. Are you having issues with the switch working for you?

Felixmil:
Switch works, but not FnKeys. With this PR, they do.

PR #1574, opened 2026-07-07
Make the Smart Pro S Home button and switch configurable

Felixmil:
On the TrimUI Smart Pro S the top Home button did nothing (the stock OSD daemon its old handler poked is absent here) and the physical switch had no spruce-configurable behaviour. This makes both configurable through a new Button Settings category, and routes every action through spruce's own systems rather than the stock scripts, whose control numbering is wrong on this device. It also fixes the in-game RetroArch menu toggle, which the device's RetroArch build never acted on.

## Button Settings

A new Button Settings category (gated to the Smart Pro S) exposes two options. Home button action picks what a Home press does: Game Switcher, Emulator menu, Exit game, Toggle LED, Screenshot, or Nothing. Switch action picks what flipping the physical switch does: LED off, Quiet Mode, Silent Mode, or Nothing. The switch selection is applied at boot from `device_init` and re-applied as the setting's `changeCmd` (with a short delay so it reads the value after PyUI has saved it).

## Home button

`device_home_button_pressed` dispatches the configured action. The old handler only did `touch /tmp/show_osdd`, a no-op here because `trimui_osdd` is not running on this device. The Emulator menu action reuses `send_menu_button_to_retroarch` so it opens the menu with the same L3+R3 combo as the Menu button (see below), while keeping the PPSSPP and pcsx signal paths that those emulators expect.

## In-game RetroArch menu

The generic 64-bit path toggles the RetroArch menu by sending `MENU_TOGGLE` over RetroArch's UDP command interface. On this device's `ra64.universal` build that socket is bound but never acts on commands, so a Menu tap did nothing in game. `send_menu_button_to_retroarch` is overridden to inject the L3+R3 controller combo via `sendevent` instead, matching `input_menu_toggle_gamepad_combo` in the device's RetroArch config, which does open the menu.

## Switch actions

All three switch actions are installed from spruce-native scripts under `spruce/smartpros/bin/`, rather than copied from the stock `fn_editor` scripts, because the stock scripts target sinks that are dead on this device. `scene-rgb-led.sh` drives the LEDs through `rgb_led_trimui` (the stock `ledc.sh` derives brightness from a shared-memory var spruce never populates, so it always resolves to dark). `scene-quiet.sh` and `scene-silent.sh` go through spruce's own `set_volume`: the stock quiet script poked `tinymix` control 9, which on this hardware is `DACL DACR Swap` (an on/off enum), not a volume control, so it never attenuated anything and could swap the audio channels. Quiet Mode now saves the current volume, drops to a low audible level, and restores it on exit; Silent Mode drops the volume to 0 (so the in-UI volume reading reflects the mute) and keeps the `/sys` speaker hard-mute as a backstop.

## Validation

Validated on a physical Smart Pro S: the switch's Quiet Mode drops both the stored `.vol` and the real DAC hardware volume (for example `.vol` 12 to 4, DAC volume 53/63 to 31/63) and restores the exact prior value on exit, with the saved-volume file cleaned up; `apply-switch-action --now` clears any previously installed scene script and installs the configured one. The L3+R3 menu combo matches the device's RetroArch `input_menu_toggle_gamepad_combo`. No `keymon` is added; the boot blob list is unchanged; both new config options are gated to `TRIMUI_SMART_PRO_S` and hidden on every other device.


spruce/scripts/platform/device_functions/SmartProS.sh
```diff
@@ -9,6 +9,27 @@
...
+# The generic 64-bit implementation (common64bit.sh) toggles the RA menu by
+# sending "MENU_TOGGLE" to RetroArch's UDP command interface (127.0.0.1:55355).
+# On this device's ra64.universal build that socket is bound but never acts on
+# commands (VERSION/GET_STATUS get no reply, MENU_TOGGLE is a no-op), so a short
+# MENU tap did nothing while a game was running; only MENU+X (RetroArch's own
+# input hotkey layer) worked. Override it to inject the configured menu-toggle
+# controller combo instead. input_menu_toggle_gamepad_combo = "2" is L3 + R3,
+# and send_virtual_key_L3R3 (from common64bit.sh) drives exactly that combo via
+# sendevent on the input device, which does open the menu.
+send_menu_button_to_retroarch() {
```
chrisj951: @Sundownersport  / @ryanmsartor  have either of you had this issue? The menu button works fine for me on SmartProS and im wondering if it's a firmware issue?

Saves/spruce/spruce-config.json
```diff
@@ -311,6 +311,35 @@
...
+                "devices": ["TRIMUI_SMART_PRO_S"],
+                "display": "Switch action",
+                "description": "What flipping the physical switch does",
+                "options": [
+                    "LED off",
+                    "Quiet Mode",
+                    "Silent Mode",
+                    "Nothing"
+                ],
+                "changeCmd": "sh /mnt/SDCARD/spruce/smartpros/bin/apply-switch-action",
```
chrisj951: Did this not work without the sh in front?

spruce/scripts/platform/device_functions/SmartProS.sh
```diff
@@ -414,7 +439,87 @@ device_prepare_for_poweroff() {
 }
 
 device_home_button_pressed() {
-    touch /tmp/show_osdd
+    # The Smart Pro S has no Fn keys, only the top Home button (KEY_HOMEPAGE) and
+    # the switch. The stock OSD daemon (trimui_osdd) is absent on this device, so
+    # the old "touch /tmp/show_osdd" was a no-op. Instead run the action the user
+    # picked in Settings -> Button Settings -> Home button action.
+    action="$(get_config_value '.menuOptions."Button Settings".homeAction.selected' "Game Switcher")"
```
chrisj951, line 441: Instead of duplicating these can you put them in a common file to be sourced?

Saves/spruce/spruce-config.json
Felixmil: Good catch. The `sh` prefix was only there because the script had lost its executable bit (`100644`), unlike the sibling `update-thermal-watchdog-to-setting` changeCmd next door, which runs directly because it is `100755`. Fixed in dd58908: `apply-switch-action` (and `scene-rgb-led.sh`) are now `+x` and both call sites (this `changeCmd` and the `device_init --now` call) invoke the script directly with no `sh` prefix.

spruce/scripts/platform/device_functions/SmartProS.sh
Felixmil: Done in dd58908. The duplicated colour-name -> hex table now lives in a single `led_color_hex` helper in `trimui_delegate.sh` (already sourced by both call sites), and both the Home "Toggle LED" action here and `scene-rgb-led.sh` call it instead of carrying their own copy.

chrisj951: I was thinking this whole function could basically be moved.

Grab the action and pass it to a function that does it. Both call the same common function

spruce/scripts/platform/device_functions/SmartProS.sh
```diff
@@ -414,7 +439,83 @@ device_prepare_for_poweroff() {
...
+            ;;
+        "Nothing"|*)
+            log_message "device_home_button_pressed: action '$action', nothing to do"
+            ;;
+    esac
+}
+
+# Self-contained helpers for device_home_button_pressed (mirrors the dispatch in
+# homebutton_watchdog.sh, which only listens for the Menu key, not Home).
+home_button_kill_emulator() {
```
chrisj951: Same with all of this and below that was copied from other functions

spruce/scripts/platform/device_functions/SmartProS.sh
Felixmil: Done in 6fb26ae. You were right that the whole thing should be shared, not just the colour table. The catch was that `perform_action` / `kill_emulator` / `prepare_game_switcher` lived inside `homebutton_watchdog.sh` (the Menu-button watchdog process), while the Home button is dispatched from a separate process (`buttons_watchdog.sh` -> `device_home_button_pressed`), so they were not reachable to call.

So I moved that dispatch into a new sourced `spruce/scripts/button_actions.sh`. Both watchdogs source it now. `device_home_button_pressed` just resolves `homeAction` and calls `perform_action "$action"` (plus the same game-state cleanup the hold-home path does). The three `home_button_*` copies are gone, and `perform_action` gained the `Toggle LED` / `Screenshot` / `Nothing` cases the Home menu adds. The moved functions are otherwise byte-identical to before, so the Menu button behaviour on every device is unchanged.

spruce/scripts/platform/device_functions/SmartProS.sh
Felixmil: Fixed in 6fb26ae. These three helpers (the emulator-kill dispatch, the RA-menu dispatch, and the game-switcher capture/kill) were copies of `kill_emulator` / `perform_action` / `prepare_game_switcher` from `homebutton_watchdog.sh`. They are all gone now: that logic moved into the shared `spruce/scripts/button_actions.sh` and the Home button calls the same `perform_action` the Menu button uses. As a bonus my copies had narrowed the process-name lists to Smart-Pro-S-only names; the shared versions carry the full cross-device lists, so converging on them is also slightly more correct.

spruce/scripts/button_actions.sh
```diff
@@ -0,0 +1,293 @@
...
+        return 0
+    fi
+
+}
+
+# Toggle the RGB LEDs between the configured colour and off, through spruce's own
+# LED system. "Off" is a static 000000 rather than effect=0, which only freezes
+# the animation on its last colour. Reads the left LED to decide which way to
+# flip. Devices without RGB LEDs never offer this action, so it stays here.
+toggle_led() {
```
chrisj951: Move this into device_functions (with a log like all the others) and this implementation of it into the SmartProS.sh file

spruce/scripts/platform/device_functions/SmartProS.sh
chrisj951: Probably just remove this and have it be a separate PR if needed in the future. It's unrelated to the current PR as is and i think something else is going wrong preventing your network loopback from working

chrisj951 (changes requested):
Mainly good with the PR just the last 2 things

spruce/scripts/button_actions.sh
Felixmil: Done in 74f56fa. `toggle_led` now follows the delegate pattern: a default stub in `device.sh` (`log_message "Missing toggle_led function" -v`, like the others) and the real `/sys/class/led_anim`-based implementation in `SmartProS.sh`. `button_actions.sh` keeps only the `"Toggle LED") toggle_led ;;` dispatch, so the device-specific LED paths no longer live in the shared file.

spruce/scripts/platform/device_functions/SmartProS.sh
Felixmil: Removed in 74f56fa. The `send_menu_button_to_retroarch` override and its comment block are gone; the "Emulator menu" action now uses the generic `common64bit.sh` UDP path like every other 64-bit device. Agreed it is out of scope here, and the loopback angle is worth investigating on its own before deciding whether any override is warranted. I will look into that separately and open a follow-up PR only if it turns out something device-side really is needed.

pablovacatello:
Hey, I was testing this new feaure in the nightly, and it seems that when I set the home button to launch game switcher, it works in game but it wont work from the main menu. Unless I'm missing something...

chrisj951:
> Hey, I was testing this new feaure in the nightly, and it seems that when I set the home button to launch game switcher, it works in game but it wont work from the main menu. Unless I'm missing something...

@Felixmil can probably allow it for the home button since we don't use it for anything in the main menu. It's restricted on the menu button due to menu having functionality there

PR #1584, opened 2026-07-27
Let the Smart Pro S Home button work from the main menu

Felixmil:
Follow-up to https://github.com/spruceUI/spruceOS/pull/1574#issuecomment-5076372874, where pablovacatello reported that with the Smart Pro S Home button set to Game Switcher it works in game but does nothing from the main menu, and https://github.com/spruceUI/spruceOS/pull/1574#issuecomment-5078017921, where chrisj951 confirmed the Home button can be allowed there since the main menu does not use it (the Menu button stays restricted because it already has a job there). Only the Game Switcher action was actually broken: Toggle LED and Screenshot already worked from the menu, and Emulator menu and Exit game are no-ops with no game running.

## Root cause

The shell asks PyUI to open the Game Switcher by touching `App/PyUI/pyui_gs_trigger`, but `MainMenu.check_for_gameswitcher()` only runs once at PyUI startup, ahead of the menu's infinite loops. In game that is enough, because ending the game restarts PyUI and the next process consumes the file. With MainUI already on screen nothing reads it, so the press is silently dropped.

## Approach

PyUI handles the Home button itself, so there is no trigger file involved and nothing to poll. `KEY_HOMEPAGE` (172) already arrives on `/dev/input/event4`, the same device `KeyWatcherController` reads on this hardware, it just had no entry in the key mapping and was discarded. Mapping it to a new `ControllerInput.HOME` makes it visible, and `get_input` consumes it next to the existing held-Menu handling. No view binds `HOME`, so it never propagates further.

The shell is untouched by this PR.

## Why it checks the setting rather than always opening the switcher

The watchdog still runs the configured Home action, and every action except Game Switcher already works from the menu. If PyUI opened the switcher unconditionally, a Home press configured as Screenshot would take the screenshot in the shell and open the switcher in PyUI at the same time. So `home_button_pressed` opens the switcher only when Game Switcher is the configured action, and leaves the rest to the watchdog. The lookup is an in-memory read of the config PyUI already holds for the settings menus, so it costs nothing per press.

## Notes

The switcher opens regardless of the `allowPyUiGameSwitcher` toggle that the held-Menu path checks, matching what the in-game and boot paths already do, so Home behaves the same in the menu as it does in game. Re-entry while the switcher is open is blocked by the existing `gs_triggered` flag. The mapping lives in the shared Trim/Miyoo provider, but no other device using it has a Home button, so it is inert everywhere else.

### DIFF
--- a/App/PyUI/main-ui/controller/controller.py
+++ b/App/PyUI/main-ui/controller/controller.py
@@ -1,6 +1,4 @@
 import time
-from pathlib import Path
-
 from controller.controller_inputs import ControllerInput
 from devices.device import Device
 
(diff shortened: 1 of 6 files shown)


chrisj951 (commented):
I think checking the sdcard each controller scan will be too much.

My original thought was just to stop PyUI and restart it.

If adding it into PyUI I think it might be okay to just always have home open the gameswitcher (can be done the same way we do check for hotkey).

Felixmil:
Thanks, that pushed this somewhere much better. Dropped the polling entirely and went with your PyUI suggestion, so the SD card is no longer touched on the controller scan.

It turns out PyUI can just see the button. `KEY_HOMEPAGE` (172) already arrives on `/dev/input/event4`, the same device `KeyWatcherController` reads on this hardware, it simply had no entry in `MiyooTrimKeyMappingProvider` and was being discarded. Mapping it to a new `ControllerInput.HOME` and consuming it in `get_input`, next to the existing held-Menu handling, means no trigger file and nothing to poll. The shell side of the previous version is reverted, so this PR is now three Python files and about 30 lines.

On always opening the gameswitcher: I went with checking the setting instead, because the watchdog still runs the configured Home action and every action other than Game Switcher already works from the menu. If PyUI opened the switcher unconditionally, a Home press configured as Screenshot would take the screenshot in the shell and open the switcher in PyUI at the same time. The check is an in-memory read of the config PyUI already holds for the settings menus, so it is not a per-press file read. Happy to make it unconditional if you would rather the setting only applied in game, but that double-fire is the reason I did not.

I also skipped the stop-and-restart approach since a full PyUI relaunch to open a menu felt heavy next to letting PyUI read the key it was already receiving.

Not yet tested on hardware, so a sanity check from someone with a Smart Pro S in hand would be welcome.

App/PyUI/main-ui/controller/controller.py
```diff
@@ -200,7 +221,12 @@ def get_input(timeout=-2, called_from_check_for_hotkey=False):
 
                 if Controller.last_controller_input is not None:
                     Theme.controller_button_pressed(Controller.last_controller_input)
-                    if Controller.last_controller_input == ControllerInput.MENU:
+                    if Controller.last_controller_input == ControllerInput.HOME:
+                        # Consumed here, never handed to a view: no menu binds it.
+                        if not called_from_check_for_hotkey:
+                            Controller.home_button_pressed()
+                        Controller.clear_last_input()
```
chrisj951, line 228: Might need to change this in the future but should be fine for now

PR #1598, opened 2026-08-10
change miyoo screenshot shortcut to menu+select

nroj95:
adds MENU+SELECT as an additional global screenshot shortcut option.

changes:
- add MENU+SELECT to the existing global screenshot shortcut setting
- fix Miyoo Mini select button mapping so B_SELECT matches the actual B_SEL input event

tested on Miyoo Mini Plus:
- screenshot from spruce ui
- screenshot in-game

### DIFF
--- a/App/PyUI/main-ui/controller/controller.py
+++ b/App/PyUI/main-ui/controller/controller.py
@@ -1,3 +1,4 @@
+import os
 import time
 from controller.controller_inputs import ControllerInput
 from devices.device import Device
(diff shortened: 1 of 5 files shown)


App/PyUI/main-ui/controller/controller.py
```diff
@@ -266,7 +267,16 @@ def get_input(timeout=-2, called_from_check_for_hotkey=False):
                             Controller.home_button_pressed()
                         Controller.clear_last_input()
                     elif Controller.last_controller_input == ControllerInput.MENU:
-                        if not Controller.is_check_for_hotkey and not called_from_check_for_hotkey and not Controller.check_for_hotkey():
+                        if not called_from_check_for_hotkey and os.path.exists("/tmp/screenshot_menu_consumed"):
```
chrisj951, line 270: There's some other code creating problems that this is just circumventing but not addressing. If you have time the update should be

Add a new key to App/PyUI/py-ui-config.json of ignoreMenuCombinationPresses

In Controller.py if ignoreMenuCombinationPresses is true, then have check_for_hotkey skip the actual performance of the hotkey, but report back that it WAS a hotkey press

This should prevent the rest of the code from seeing the menu, and allow delegating the hotkey solely to outside of PyUIs control instead of it being a weird shared handoff

chrisj951 (changes requested):

chrisj951:
If you remove the PyUI side of this I am okay with this PR

PR #1607, opened 2026-08-22
Anbernic XX: remove stockOS compatibility, target BaseOS only

ryanmsartor:
Removes stockOS compatibility from the Anbernic H700 (XX) line, so those devices target BaseOS only.

## What this drops

- Every `if [ -n "$SPRUCE_BASEOS" ] … else <stock> fi` fork across the XX line — apps, emulator launchers, the platform cfg and device functions.
- The `lcd_type` + `/mnt/vendor/oem/board.ini` device probe in `helperFunctions.sh`. BaseOS names the target exactly in `/etc/baseos-release`, and stock can no longer boot spruce here, so the fallback only produced a device profile that was wrong in every field. `SPRUCE_BASEOS` is now set unconditionally for the H700 line.
- Stock-only files: `spruce/rg34xxsp/loadapp.sh`, an unreferenced `spruce/rg35xxsp/ANBERNIC-keys.cfg`.
- Three `systemctl` calls from `.tmp_update/anbernic.sh` — that file is now reached only from the BaseOS hand-off shim, and BusyBox has no systemctl.
- `/mnt/vendor/bin` from `PATH`; `log_dir` / `xmb_font` in `retroarch-AnbernicRG_XX-universal.cfg` that pointed into the vendor partition (now the values every other platform uses); dead vendor-partition notes in `setup_for_retroarch` and the PyUI device classes.
- `DEVICE_PYTHON3_PATH` is named outright rather than preferring a system python that both cannot exist here and would lack spruce's site-packages if it did.

`spruce/rg28xx/8188eu.ko` moves to `spruce/h700/rg28xx/`, alongside `lib32`/`lib64`.

## Two bugs fixed along the way

**PPSSPP would have regressed on five other devices.** `run_ppsspp()` is the *shared* PSP launcher. Removing the gate made `export SDL_VIDEODRIVER=mali` unconditional, so A30, Flip, MiyooMini, Pixel2 and SmartProS would all have been told to use a video driver they do not have. It is now folded into the existing per-platform `case`, beside the PowerVR arm. (The `export_sdl_gamecontroller_map` call next to it was already harmless elsewhere — it returns early when `SDL_GAMECONTROLLER_MAP` is unset.)

**`AnbernicRG28XX.sh` had drifted.** It overrides `device_init`, so it never picked up anything the common `device_init` grew: no `/bin/bash` provisioning — which silently kills every watchdog and with it the power and home buttons — no `add_spruce_system_user`, and no `shield_baseos_session`, leaving its shutdown unmount unprotected. It also still launched two paths that do not exist in the tree: `/mnt/SDCARD/anbernic_adbd/run_adbd.sh` (which the common path deliberately dropped, because BaseOS binds its own adb gadget) and `spruce/rg34xxsp/bin/joypad_shim`, deleted long ago. The shared body is now `anbernic_xx_common_init()`; the RG28XX calls it and keeps only its WiFi module load.

## Notes for review

- Where `BASEOS_TARGET` comes back empty, detection falls through to a real profile rather than leaving `PLATFORM` unset — an unset `PLATFORM` fails at `. $PLATFORM.cfg` with no clue why.
- The one remaining `SPRUCE_BASEOS` test, in `ra_functions.sh`, is intentional: `apply_baseos_ra_overlay()` is called from shared code on every platform, so it still needs the gate.
- Existing cards keep a stale `spruce/rg28xx/8188eu.ko` after upgrading. It is inert — nothing loads from the old path now.
- No behaviour change intended beyond the two fixes above.

## Testing

Not yet hardware-verified. Static checks only: all touched shell files pass `sh -n`, PyUI files parse, and the branch merges cleanly into Development. Worth a boot on a CubeXX and, if anyone has one, an RG28XX — that is where the device_init fix lands.


Sundownersport:
## RG28XX testing checklist

@ryanmsartor — you're the only one with an RG28XX, and I don't think spruce-on-BaseOS has ever booted on one. So this is closer to first light than to a regression check.

### Why I think it's never been booted

`AnbernicRG28XX.sh` overrides `device_init`, so it never ran the common one — which means it never got `/bin/bash` installed. `launch_startup_watchdogs` runs *every* watchdog through `/bin/bash`, and BaseOS is BusyBox with none. On that device the power button and home button would both have been completely dead, silently. Combined with the two paths it was launching that don't exist anywhere in the repo (`/mnt/SDCARD/anbernic_adbd/run_adbd.sh` and `spruce/rg34xxsp/bin/joypad_shim`), the file reads like it was written for stock and never revisited.

### What this PR changes for the 28xx

- [ ] **Boots to PyUI**
- [ ] **Power button** sleeps and wakes (was dead — no `/bin/bash`)
- [ ] **Home button** opens the menu (same cause)
- [ ] `/bin/bash` exists after boot
- [ ] **Shutdown, then reboot** — no `Volume was not properly unmounted` in the boot log. That's `shield_baseos_session`, which this device never called before.
- [ ] **WiFi** — the module moved to `spruce/h700/rg28xx/8188eu.ko`. This is the one with a real chance of failing on its own; nobody has watched it load. Check `Saves/spruce/rg28xx_8188eu_log.txt` (insmod errors now land there — the redirect was stdout-only before) and that `wlan0` shows up.

### Pre-existing, and this is the device that can settle it

**Rotation on the 28xx is decided in four places and they disagree.** Not caused by this PR:

| Where | Says |
|---|---|
| `anbernic_rg28xx.py` `screen_rotation()` | `0` |
| `AnbernicRG28XX.cfg` `DISPLAY_ROTATION` | `270` |
| `ra_functions.sh` (hardcoded for this model) | `video_rotation = 1`, forced 640x480 |
| `pico8_functions.sh` (keyed off `DISPLAY_ROTATION`) | `transform_screen 135` |

You set the PyUI `0` in `280702b0d` fixing a real user report, and noted it'd probably be wrong on stock — stock is gone now, so that part's settled. But nothing else followed. If `0` is right, RetroArch and PICO-8 are both rendering sideways.

It's not automatically three-against-one: PyUI and RetroArch reach the panel through different stacks, so one seeing a pre-rotated surface and the other not is physically possible. Which is why it needs eyes rather than a patch.

The check is just **which of these are upright**:

- [ ] Boot logo
- [ ] PyUI
- [ ] RetroArch (any game)
- [ ] PICO-8
- [ ] Video player

Whichever is sideways names the wrong number.

### Also unverified, if you have patience

The 28xx is in the **stickless** control group, and those maps are MustardOS-derived rather than measured. The 32-bit RetroArch case is weakest: `write_baseos_ra_autoconfig_linuxraw`'s stickless *button* indices are derived by reasoning about joydev ordering, never measured. (The d-pad axes in that same block were later corrected by measurement on an RG SP — which is how we know that reasoning can be wrong.)

- [ ] RetroArch 64-bit — all buttons, incl. L2/R2, d-pad, Select/Start, hotkey combos
- [ ] RetroArch 32-bit (`ra32.h700`, linuxraw) — same
- [ ] PPSSPP controls
- [ ] PICO-8 controls

Anything in the last two sections that fails is a separate fix, not a blocker for this PR.

PR #1614, opened 2026-08-25
Wifi switch

zenkalia:
Make the physical switch on the trimui smart pro s function like wifi switch on a real PSP!

Questions:
1. What about the TrimUI Smart Pro? I should probably mirror these changes to `App/PyUI/main-ui/devices/trimui/trim_ui_smart_pro.py`, yeah?
2. Is this going to cause trouble for any other platforms? Does spruce support any hardware that _has_ a switch but _does not have_ wifi?

### DIFF
--- a/App/PyUI/main-ui/devices/trimui/trim_ui_device.py
+++ b/App/PyUI/main-ui/devices/trimui/trim_ui_device.py
@@ -32,10 +32,22 @@ class TrimUIDevice(DeviceCommon):
 
     def on_system_config_changed(self):
         old_volume = self.system_config.get_volume()
+        old_wifi_enabled = self.system_config.is_wifi_enabled()
         self.system_config.reload_config()
         new_volume = self.system_config.get_volume()
         if(old_volume != new_volume):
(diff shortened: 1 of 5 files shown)


ryanmsartor:
The Smart Pro, Smart Pro S, Brick, and Brick Pro all have this function switch, and all have WiFi. No other spruce devices have an exposed dip switch like this. I would mirror the changes for each of these 4 devices.

Sundownersport:
Looking into this PR now, likely to have some significant changes since it kind of kicked me into gear to fully implement the FN switch options menus for the TrimUI devices.

zenkalia:
I wouldn't be surprised if this had issues. I will say, along with those other two bug fixes... this does work and it's pretty glorious. I was shocked how small the diff was. 

One thing I didn't test was flipping the switch back and forth many times very quickly.

My other concern was a possible performance impact. 

Also my claude code expires sept 10 so LMK how I can help. I could vibe code some benchmark tooling to understand performance impacts. 🤷

PR #1619, opened 2026-08-27
Fix/ota full archive excludes diff

ryanmsartor:


App/-OTA/downloader.sh
```diff
@@ -296,7 +322,14 @@ if [ -n "$NIGHTLY_DIFF_LINK" ] && [ -n "$NIGHTLY_DIFF_CHECKSUM" ] && [ -n "$NIGH
     NIGHTLY_DIFF_OK=1
 fi
 
-log_message "OTA: Installed nightly base: ${INSTALLED_NIGHTLY_BASE:-none}; nightly diff base: ${NIGHTLY_DIFF_BASE_VERSION:-none} (usable: $NIGHTLY_DIFF_OK)"
+log_message "OTA: Installed nightly base: ${INSTALLED_NIGHTLY_BASE:-none}; installed nightly version: ${INSTALLED_NIGHTLY_VERSION:-none}; nightly diff base: ${NIGHTLY_DIFF_BASE_VERSION:-none} (usable: $NIGHTLY_DIFF_OK)"
+
+# A nightly that knows its own version is not offered the same (or an older)
+# nightly again.
+if [ "$TARGET_CHANNEL" = "nightly" ] && [ "$SKIP_VERSION_CHECK" != "True" ] && [ -n "$INSTALLED_NIGHTLY_VERSION" ] && nightly_is_current; then
```
ryanmsartor, line 329: You should permit same-nightly-version reupdating. This makes for easier testing of the OTA feature, for one thing, and also more gracefully handles the not-uncommon case where we generate multiple nightly versions on the same date.

ryanmsartor (commented):
one change requested, to allow OTA of nightlies even when apparently the same date, since the same date may have multiple "nightly" builds spun in its name; silly GH won't let me put this as a change request because I'm the one who opened the PR.

PR #1623, opened 2026-08-28
Add a home button option for chorded hotkeys

zenkalia:
Adds "Hotkeys" as an option under

Settings > Additional Settings > Emulator Settings > Hold home key action

When this option is selected, instead of triggering a behavior when held for 3s, make the menu button into a button that you can chord with other buttons for different behavior:

menu + vol_up = brightness up
menu + vol_down = brightness down
menu + A = reset rom
menu + B = emulator menu
menu + R = save state in the currently selected save slot 
menu + L = load state from the currently selected save slot 
menu + Start = Exit game

TL;DR: Knulli hotkey mode.


zenkalia:
Must remember to update the wiki if/when this gets merged.

ryanmsartor:
Hotkeys are already hooked up in their relevant emulators, and can already be chorded with menu if that's what you set your hotkey enable key to. Homebutton watchdog already considers this, and will cancel out of the long press action if another key is pressed.

I like the brightness hotkeys proposed here, but I would rather that be its own independent setting from anything emulator-related.

PR #1624, opened 2026-08-28
Sync Script

zenkalia:
This has been my one liner that speeds up device testing.

Usage:

`bash sync-to-device.sh /Volumes/SDCARD/`

Once you have sync'd with the device you can `git add` those changes. By design, this only pushed unstaged changes.

Edit -> Sync -> Git Add -> Test -> REPEAT

If we don't want something like this in the repo, maybe we should have a tools or scripts folder in .gitignore instead.


Sundownersport:
This is sort of our dumping ground for things like this https://github.com/spruceUI/spruceSource

PR #1625, opened 2026-08-29
Add the option to use the home button to control brightness.

zenkalia:
This chord is also available on the home screen and we disable the little top left menu when the option is selected.


Sundownersport:
I’m just wrapping up my rgb30 work for tonight but I’ll take a look at this tomorrow

spruce/scripts/buttons_watchdog.sh
```diff
@@ -148,12 +176,12 @@ getevent $EVENTS | while read line; do
...
         ;;
         *"key $B_L1 1"*) # L1 key down
-            if [ "$START_DOWN" = true ] ; then
+            if [ "$BRIGHTNESS_HOTKEY" = "Start + L/R" ] && [ "$START_DOWN" = true ] ; then
                 brightness_down
             fi
         ;;
         *"key $B_R1 1"*) # R1 key down
-            if [ "$START_DOWN" = true ] ; then
+            if [ "$BRIGHTNESS_HOTKEY" = "Start + L/R" ] && [ "$START_DOWN" = true ] ; then
```
zenkalia, line 184: I feel like `BRIGHTNESS_HOTKEY` should be holding a boolean instead of the user option but maybe I'm just playing code golf.

Saves/spruce/spruce-config.json
```diff
@@ -394,6 +394,7 @@
                     "Emulator menu",
                     "Game Switcher",
                     "Exit game",
+                    "Brightness (Menu + Vol)",
```
ryanmsartor, line 397: Can we have the Brightness hotkeys be its own entry instead of lumping it in with the hold entries? I don't think menu + volume for brightness needs to be mutually exclusive with the others.

On the sh side at least, we already have the capability to cancel out of the long press action if another button is pressed - I would want to leverage this to allow menu to be further overloaded.

spruce/scripts/buttons_watchdog.sh
```diff
@@ -108,6 +133,9 @@ take_screenshot_bg() {
     take_screenshot "$ss_name"
 }
 
+# Setup brightness hotkey (Button Settings > Brightness hotkey)
+BRIGHTNESS_HOTKEY="$(get_config_value '.menuOptions."Button Settings".brightnessHotkey.selected' "Start + L/R")"
```
ryanmsartor, line 137: Here you look to be keying off brightnessHotkey whereas in the config itself you had modified the holdHomeAction section. Did you include the wrong version of the config in this PR perhaps?

spruce/scripts/buttons_watchdog.sh
ryanmsartor: I think it's probably better to leave it a string so that we can just pull it straight from the config instead of having an intermediate step - but it probably doesn't really matter either way to be honest. Dealer's choice!

spruce/scripts/homebutton_watchdog.sh
```diff
@@ -32,33 +32,46 @@ home_key_down () {
...
-                log_message "homebutton_watchdog.sh: Performing hold-home action: $HOLD_HOME"
-                perform_action "$HOLD_HOME"
-
-                # Only clean up game state when the action actually killed the game.
-                # "Emulator menu" just opens an in-game menu — the game is still running.
-                if [ "$HOLD_HOME" != "Emulator menu" ]; then
-                    rm -f /tmp/cmd_to_run.sh
-                    rm -f /mnt/SDCARD/spruce/flags/lastgame.lock
-                fi
+        HOLD_HOME="$(get_config_value '.menuOptions."Emulator Settings".holdHomeAction.selected' "Game Switcher")"
```
ryanmsartor, line 35: Here again you're pulling from holdHomeAction. Again I think it would be better to swap everything to a dedicated brightnessHotkey entry.

ryanmsartor (changes requested):

Sundownersport:
Took a pass at this on `feat/brightness-hotkey` (branched off current Development, @zenkalia's commit cherry-picked intact with a rework commit on top so the delta is readable). Landing where Ry did on #1 — dedicated setting, not a hold-home action.

**The exclusivity wasn't needed in the first place.** `cancel_menu_hold()` in `homebutton_watchdog.sh` already lists the volume keys among the keys that cancel a MENU hold, and `home_key_up` already skips the tap-home action when `was_cancelled` is set. So holding MENU and pressing a volume key kills the hold timer and suppresses the tap on its own — there's nothing left for the chord to be exclusive *with*. The `"Brightness (Menu + Vol)"` branch in `home_key_down` is gone and that function is byte-identical to Development again.

**In its place:** `Button Settings > Brightness hotkey` — `Start + L/R` / `Menu + Vol` / `Both` / `Off`, no `devices` restriction. Ry's third comment (the `brightnessHotkey` vs `holdHomeAction` mismatch) wasn't a wrong config file, it was this setting half-landed: `buttons_watchdog.sh` reads `"Button Settings".brightnessHotkey` but the key was never added to `spruce-config.json`, so that gate always fell through to its default. Now it has an entry behind it. Kept as a string per Ry — with `Both` in there it was never going to be a boolean anyway.

Read at press time rather than cached at startup so a change applies without a reboot; the cheap test (`[ -e /tmp/menubtn ]`, `$START_DOWN`) runs first so jq only forks on a press that could actually be a chord.

**Defaults to `Both`**, so nothing loses a chord it had. Worth a second opinion: the cost is that in game, holding Menu and tapping volume now changes brightness instead of volume. That gesture already cancelled your hold-home action so it wasn't doing much, but it is a default behavior change — `Start + L/R` would be the conservative pick if you'd rather this be opt-in.

**Upgrades need no script.** `merge_configs.py` drops a `selected` value that's no longer in the new option list, so anyone who tested this build with `holdHomeAction = "Brightness (Menu + Vol)"` falls back to `Game Switcher` and picks up `brightnessHotkey = Both`.

**Kept from the original:** the `EVENT_PATH_VOLUME` addition in `homebutton_watchdog.sh`. On SmartPro/SmartProS the volume keys report on their own node, so that watchdog never saw them and `cancel_menu_hold()` never fired — which is a standalone bug (a vol press while MENU is held lets the hold-home action fire out from under it) *and* the thing that makes this chord work there at all. Worth landing on its own merits.

**PyUI side:** reverted the `main_menu.py` change. Reading the config to refuse to open the popup menu costs you the popup permanently in that mode. The real defect is that `check_for_hotkey()` only watches for 300ms, so a vol press landing mid-hold missed the window and MENU fell through to its own action. Fixed in `controller.py`: where nothing else claims a held MENU (game switcher off — which is the case that produces the bug), keep watching for the chord as long as MENU is held. `check_for_hotkey()` also honors the new setting now, so `Off` actually turns it off in the UI instead of swallowing MENU and doing nothing. Note PyUI has done Menu+Vol brightness unconditionally on every screen for a while (`perform_hotkey`) — this just makes it respect the setting and stop missing late presses.

**Not hardware-tested yet.** Two things I'd want on a device before this merges:

- The chord window is your Menu Hold Time, and the shipped default is **1 second** — hold Menu a full second before reaching for volume and GS opens first. It degrades cleanly (the timer `rm`s `/tmp/menubtn`, so once GS is up the chord is off and volume is volume again), but whether 1s is comfortable in practice isn't something I can tell from the code. There's a real tension: raise it for the chord, slow down GS.
- The `EVENT_PATH_VOLUME` path specifically wants a SmartPro/SmartProS.

Saves/spruce/spruce-config.json
zenkalia: Having tested both ways... I do think they should be mutually exclusive. Config sprawl is already pretty bad in SpruceOS and most all of the handhelds Spruce supports have both a menu button and a home button that are both configurable.

I won't die on this hill, though. I can move it to a new config entry.

zenkalia: Ah nevermind sundowner beat me to it!

Sundownersport:
Merged as #1628 — your commit went in as-is with the review rework on top, so the authorship is intact in the history. Closing this one out. Thanks for the feature, and for the digging on the SmartPro volume node in particular — that turned out to be a real standalone bug in `cancel_menu_hold`, independent of the brightness chord.

Still wants a hardware pass (notes in #1628), so shout if you hit anything odd on your device.

PR #1633, opened 2026-08-31
Batch fix automated

CatalyticArkun:

PR #1634, opened 2026-08-31
Fix five regressions from the automated batch (#1633), and two bugs it surfaced

Sundownersport:
PR #1633 traded several working oddities for clean-looking breakage. Five of these are regressions from that batch; two are older bugs found while reviewing it.

The recurring theme: #1633 removed things that looked redundant but weren't — a path detour, an `eval`, an uppercase file extension.

### RandomGame launched every game with no emulator
`standard_launch.sh:22` takes its identity from `$0` — it strips up to `/Emu/` and keeps the next path component as `EMU_NAME`. The `Emu/<system>/../../` detour was load-bearing, not redundant. The canonical path contains no `/Emu/`, so `EMU_NAME` came out empty and `EMU_JSON_PATH` became `/mnt/SDCARD/Emu//config.json`.

Three consumers broke:
1. Core/emulator selection
2. The game switcher — `button_actions.sh` greps `Emu/<system>/` back out of `/tmp/cmd_to_run.sh`
3. **Auto-resume** — `save_poweroff.sh:180` (`close_non_emu_cmd_to_run`) greps for `/mnt/SDCARD/Emu` to tell an emulator from an App. Every random launch classified as an App, and that branch does `rm -f "${FLAGS_DIR}/lastgame.lock"`

The `eval` removal from that commit is kept.

### Every Moonlight stream failed
`gui/main.lua:1339` writes shell syntax — quoted arguments and a literal `$GAMEDIR` — which only worked because `launch.sh` ran it through `eval`. Replacing `eval` with `set -- $COMMAND` does neither quote removal nor parameter expansion, so moonlight received the literal `"Desktop"` *with* its quotes and the literal string `$GAMEDIR/keys`. Multi-word app names ("Steam Big Picture") split into several arguments.

Fixed at the producer: one argument per line, read back with `while IFS= read -r`. `launch.sh` contributes `stream` and `-keydir` because it's what knows `GAMEDIR`.

Two guards go with it — a `nil` in a Lua table constructor is *dropped* rather than stored, so `table.concat` would quietly write a truncated argument list where the old `..` raised.

### ROM lists crashed on any gamelist entry without `<image>`
`image=None` reaches `CachedExists.exists` → `os.path.normpath(None)` → `TypeError`, raised lazily while the grid draws and outside any handler. `""` is not a safe placeholder either: `normpath("")` is `"."`, which exists.

### `EBOOT.PBP` disappeared from PSX
PyUI lowercases both sides, so dropping the uppercase entry was a no-op there. But `MiyooGamelist/generate.py:374` and `RandomGame/random.sh:44` both match case-sensitively, and uppercase is the canonical form for these images.

### "Only when unplugged" could no longer sleep on unplug
The edge trigger latched the raw lid state even on the loop where the charging check rejected the close. Closing the lid while charging and unplugging later never slept until the lid was physically cycled. Now latches whether *this close* was acted on, separately from the lid state, cleared when the lid opens.

### A `<game>` with `<path>` but no `<name>` takes down the ROM list sort
`name=None` becomes `display_name`, and `get_sort_key()` calls `.strip()` on it. `rom_select_options_builder.py:352-353` sorts every file and folder list that way. Falls back to the filename the way a missing entry already does.

### `AnbernicXXCommon.cfg` set `LED_PATH=""`
Every other LED-less platform uses the `"not applicable"` sentinel, and it isn't cosmetic: `LED_PATH` is concatenated, so an empty value resolves to `/brightness` and `/trigger` and the writes land at the filesystem root — on every boot from `runtime.sh:18`, every shutdown from `save_poweroff.sh:30`, and every 30s once `low_power_warning.sh` started running there.

## Testing

Verified by source reading and local execution (Lua, shell and Python all run on the dev machine):

- `EMU_NAME` resolves to the system again, and the game-switcher `sed` and the autoresume grep both classify correctly
- Moonlight end-to-end: real LuaJIT writes `command.txt`, the actual reader parses it — 21 arguments, multi-word app name intact as one, keydir expanded, IP trailing newline stripped
- Lid watchdog simulated against old and new logic across four timelines: old gives 0 sleeps in the close-while-charging-then-unplug case, new gives exactly 1 at the moment of unplug, with the no-repeat and re-arm properties intact
- All touched shell scripts pass `sh -n`, Python compiles, Lua loads, JSON parses
- Line endings preserved (`rom_select_options_builder.py` is CRLF)

**Not tested on hardware.** The Moonlight change touches both sides of a producer/consumer pair and deserves a real stream before merge.

## Not in this PR, but worth knowing

- `archiveUnpacker.sh` — the tightened member check is correct and should stay, but it's the first time that check has ever actually run (the old grep matched the listing header unconditionally, so it always passed). Expect "my theme stopped installing" reports.
- `button_actions.sh:322` — the new `wait` before `killall sendevent`. Couldn't verify the hang risk either way; `sendevent` ships as a binary with no source in-tree.
- The new language keys leave 149 untranslated strings per language. No existing translation was clobbered — checked all 28 files key by key.

### DIFF
--- a/App/Moonlight/launch.sh
+++ b/App/Moonlight/launch.sh
@@ -18,16 +18,24 @@ chmod +x ./moonlight/moonlight
 ./love gui
 
 cd "$MOONDIR"
-COMMAND=$(cat command.txt)
-
 export GAMEDIR
-/mnt/SDCARD/spruce/bin64/gptokeyb "moonlight" &
-set -f
-# shellcheck disable=SC2086
-set -- $COMMAND
-set +f
-./moonlight "$@"
-kill -9 $(pidof gptokeyb) 2>/dev/null
+
+# No command.txt means the GUI was closed without picking an app.
+#
+# It is written by gui/main.lua, one argument per line. Read it that way:
+# word-splitting it would neither strip quotes nor expand anything, so an app
+# name with spaces would arrive as several arguments. "stream" and -keydir go
+# on here because this is the script that knows GAMEDIR.
+if [ -f command.txt ]; then
+    set -- stream -keydir "$GAMEDIR/keys"
+    while IFS= read -r moonlight_arg || [ -n "$moonlight_arg" ]; do
+        [ -n "$moonlight_arg" ] && set -- "$@" "$moonlight_arg"
+    done < command.txt
+
+    /mnt/SDCARD/spruce/bin64/gptokeyb "moonlight" &
+    ./moonlight "$@"
+    kill -9 $(pidof gptokeyb) 2>/dev/null
+fi
 
 rm -f command.txt
 rm -f /tmp/stay_awake
--- a/App/Moonlight/ports/moonlightnew/gui/main.lua
+++ b/App/Moonlight/ports/moonlightnew/gui/main.lua
@@ -1323,6 +1323,14 @@ end
 function writeSelectedApp(selectedApp, bitrate, resolution, framerate, codec, remote)
     -- Split resolution string into width and height
     local width, height = resolution:match("(%d+)x(%d+)")
+    if not width or not height then
+        -- A nil here used to blow up on the .. concatenation below. It must
+        -- not reach the args table instead: a nil in a table constructor is
+        -- dropped rather than stored, so table.concat would quietly write a
+        -- truncated argument list and moonlight would fail for no clear reason.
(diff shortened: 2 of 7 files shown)


CatalyticArkun (approved):
🚀

PR #1637, opened 2026-09-03
DSperate follow-ups: reach kill_dsperate, and stop guessing state paths

Sundownersport:
Review follow-ups on the DSperate series (`9ff7f6734..f4a38876e`). Nothing here changes the design, it just closes gaps in it.

## The two that matter

**`kill_emulator` never reaches `kill_dsperate`.** `f4a38876e` renamed the binary inside `kill_dsperate()` but not in the branch that decides to call it:

```sh
elif pgrep dsperate-sdl >/dev/null; then   # never matches now
```

So a power or menu longpress fell through to `kill_ra_and_standard_emulators`, which doesn't list DSperate — meaning no SIGTERM, and both the battery save and the `autosave = true` state are lost. Exactly what that commit was trying to protect.

**`pin_to_dedicated_cores` is still broken.** `6f1faa1ed` quoted the `EMU_CPUS` line, but the defect is one line up and untouched:

```sh
delay="$2:-1"      # a literal, not ${2:-1}
```

`sleep :-1` / `sleep 2:-1` both fail, so the backgrounded `pgrep` runs before the emulator exists and nothing gets pinned. DSperate no longer calls this, but RetroArch (no `$2`) and DraStic (`2`) still do.

## Auto-resume path handling

`get_game_code` covered `.nds` and `.zip`, but NDS `extlist` is `nds|zip|7z|rar`. An unreadable header gave an empty code and `get_state_path` returned a bare `.auto.dss` — a path DSperate never writes, and one every game would share.

- `.7z` now goes through `7zr e -so`, `.zip` through python's `zipfile`, matching how `mupen_functions.sh` and `bigpemu_functions.sh` already open archives. `unzip` isn't in `spruce/bin` and was used nowhere else in the tree, so the old `.zip` branch was relying on the busybox applet being present.
- `get_state_path` returns empty when no code comes back, and `run_dsperate` checks before adding `--load-state`. `.rar` has no reader available either way, so it now skips the resume cleanly instead of pointing at a shared name.

Verified locally against a synthetic ROM in all four shapes:

```
game.nds   code=[ABCD] state=[.../ABCD.auto.dss]
game.zip   code=[ABCD] state=[.../ABCD.auto.dss]
game.7z    code=[ABCD] state=[.../ABCD.auto.dss]
junk.rar   code=[]     state=[]
```

## Smaller things

- `is_autoload_enabled` read a hardcoded config path with no fallback. Now uses the exported `EMU_JSON_PATH` with `// "Enabled"`. (`get_config_value` is *not* right here — it reads `spruce-config.json`, not the per-emu one.)
- `dsperateAutoLoad` was missing the `"overrides": {}` every sibling in that block has.
- BIOS paths go back on the command line. They're in `spruce.ini`, but they're also what `dsperate_bios_missing` just checked for, and a card seeded during the mid-series nightlies (after `XDG_CONFIG_HOME` moved to `/mnt/SDCARD/Saves`, before `spruce.ini` landed) has an ini with no `[paths]` block — the BIOS check passes and then DSperate can't find them. Flags override the file, so this makes the checked files the used files. The two near-identical invocations collapse into one `set --` build-up.
- `resetDSperate.sh` trusted `rm -rf`'s exit status, which is 0 whether or not anything was there, so the failure branch was dead. It now checks the directory is gone, and also clears the orphaned `Emu/NDS/config` tree from the old XDG path. Saves and states are left alone.

## Not touched

Collapsing `Emulator_BrickPro` into `Emulator_SmartPro` drops `DraStic-Steward` from the Brick Pro's options and adds `DraStic-trngaje`. That reads as intentional per `efbcd8023`, so I left it — just flagging it.

## Testing

`sh -n` clean on all four scripts, `jq -e .` clean on the config. The game-code paths are hardware-independent and tested above. The `kill_dsperate` and pinning fixes want a device check.

### DIFF
--- a/Emu/NDS/config.json
+++ b/Emu/NDS/config.json
@@ -89,7 +89,8 @@
         "Enabled",
         "Disabled"
       ],
-      "selected": "Enabled"
+      "selected": "Enabled",
+      "overrides": {}
     }
   },
   "sortOrder": 4000,
--- a/spruce/scripts/emu/lib/dsperate_functions.sh
+++ b/spruce/scripts/emu/lib/dsperate_functions.sh
@@ -57,8 +57,9 @@ is_autoload_enabled() {
 
 # The NDS header carries a four-character game code at offset 12, and DSperate
 # names the auto slot after it, so the launcher has to read it the same way to
-# know which file to resume from. 7zr cannot open a .rar and neither can
-# python's zipfile, so those come back empty rather than wrong.
+# know which file to resume from. Only the two containers DSperate itself
+# opens are handled: a .7z has already been unpacked by prepare_dsperate_rom
+# before this is called, and nothing on the card can read a .rar.
 get_game_code() {
 	case "$1" in
 	*.nds | *.NDS)
@@ -74,9 +75,6 @@ with zipfile.ZipFile(sys.argv[1]) as z:
             break
 " "$1" 2>/dev/null
 		;;
-	*.7z | *.7Z)
-		7zr e "$1" -so 2>/dev/null | dd bs=1 skip=12 count=4 2>/dev/null
-		;;
 	esac
 }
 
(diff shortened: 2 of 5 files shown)


Emu/NDS/config.json
```diff
@@ -89,7 +89,8 @@
         "Enabled",
         "Disabled"
       ],
-      "selected": "Enabled"
+      "selected": "Enabled",
+      "overrides": {}
```
ryanmsartor, line 93: this is fine, but note we don't have any handling for overrides for this item yet.

spruce/scripts/emu/lib/dsperate_functions.sh
```diff
@@ -49,31 +49,44 @@ seed_dsperate_config() {
...
 # returns 0=true or 1=false; for now, no per-game override logic in place.
 is_autoload_enabled() {
-    autoload_setting="$(jq -r '.menuOptions.dsperateAutoLoad.selected' "/mnt/SDCARD/Emu/NDS/config.json")"
-    if [ "$autoload_setting" = "Enabled" ]; then
-		return 0
-	else
-		return 1
-	fi
+	_setting="$(jq -r '.menuOptions.dsperateAutoLoad.selected // "Enabled"' "${EMU_JSON_PATH:-/mnt/SDCARD/Emu/NDS/config.json}")"
+	[ "$_setting" = "Enabled" ]
```
ryanmsartor, line 55: this change is probably unnecessary but fine, but changes the fallback behavior from disabled by default to enabled by default.

spruce/scripts/emu/lib/dsperate_functions.sh
```diff
@@ -49,31 +49,44 @@ seed_dsperate_config() {
...
+		"$(get_python_path)" -c "
+import sys, zipfile
+with zipfile.ZipFile(sys.argv[1]) as z:
+    for name in z.namelist():
+        if name.lower().endswith('.nds'):
+            sys.stdout.write(z.open(name).read(16)[12:16].decode('latin-1'))
+            break
+" "$1" 2>/dev/null
+		;;
+	*.7z | *.7Z)
```
ryanmsartor: both .7z and .rar are not supported by dsperate. Maybe we could do a mupen-like temp extraction to give support for .7z, and then leave this change in, but do we wanna consider removing .rar from extlist? Does drastic even open .rar?

re: changing unzip to python zipfile, you won me over with this in the end because A30 has no unzip bin - but we use python on every device spruce touches. Fair.

spruce/scripts/emu/lib/dsperate_functions.sh
```diff
@@ -103,18 +116,21 @@ run_dsperate() {
...
-			--fullscreen \
-			--load-state "$state_path" \
-			> "$(emu_log_file)" 2>&1
-	else
-		./dsperate "$ROM_FILE" \
-			--fullscreen \
-			> "$(emu_log_file)" 2>&1
+	# The BIOS paths are in spruce.ini too, but pass them anyway: they are what
+	# dsperate_bios_missing just checked for, and a card upgraded from a config
+	# seeded before spruce.ini existed has no [paths] block at all.
```
ryanmsartor, line 176: meh. I kinda like the idea of leaving it user-configurable, but then we'd have to change how the bios check functions. This is fine I guess.

spruce/scripts/emu/lib/general_functions.sh
```diff
@@ -185,7 +185,7 @@ set_cpu_mode() {
 
 pin_to_dedicated_cores() {
 	comm="$1"
-	delay="$2:-1"
+	delay="${2:-1}"
```
ryanmsartor, line 188: 👍

ryanmsartor (changes requested):

spruce/scripts/tasks/resetDSperate.sh
```diff
@@ -2,8 +2,14 @@
 
 . /mnt/SDCARD/spruce/scripts/helperFunctions.sh
 
-if rm -rf /mnt/SDCARD/Saves/dsperate; then
-    log_message "dsperate configs have been reset."
+# The live config is under XDG_CONFIG_HOME, which run_dsperate points at
+# /mnt/SDCARD/Saves. Emu/NDS/config is where it used to live, so clear that too
+# or an upgraded card keeps a stale copy nobody reads. Saves and states are
+# deliberately left alone.
+rm -rf /mnt/SDCARD/Saves/dsperate /mnt/SDCARD/Emu/NDS/config
```
ryanmsartor: forgot to comment on this. DO NOT NUKE EMU/NDS/CONFIG

Sundownersport: Dropped it entirely. Emu/NDS/config is DraStic's — the drastic-*.cfg files plus the .bak copies resetDraStic.sh restores from. Would have wiped both. Good catch.

spruce/scripts/emu/lib/dsperate_functions.sh
Sundownersport: Added the extraction. Two notes:

Goes to /mnt/SDCARD/spruce/tmp, not mktemp — /tmp is tmpfs and NDS ROMs are big.

Keeps the archive's filename, because save_path() in dsperate's main.cpp is `base_name(rom_stem(rom)) + ".sav"` — a fixed rom.nds would give every 7z game one shared .sav.

Don't drop .rar: `strings drastic64` has `%s unrar'ed`, plus zip and 7z. extlist is DraStic's list and right for DraStic; DSperate is just narrower.

spruce/scripts/emu/lib/dsperate_functions.sh
Sundownersport: Enabled matches the shipped `selected` value, so a missing key behaves like a fresh install. Happy to flip it if you'd rather it fail closed.

ryanmsartor (approved):

PR #1646, opened 2026-09-07
Anbernic XX: RetroArch and standalone pad mappings in parity with the arm64 fleet

CatalyticArkun:
## Summary

Gives the Anbernic RG XX line (H700, BaseOS) the same controller chords as the other arm64 spruce devices, sticks included, for RetroArch and every standalone emulator spruce ships on the line. MENU is the RetroArch modifier, as on the rest of the fleet.

**RetroArch (both builds):** MENU + B exit · + A screenshot · + X menu · + Y fps · + L1/R1 load/save · + L2/R2 slow-motion/fast-forward · + UP shader · + LEFT/RIGHT state slot. Sticks: left a0/a1, right a2/a3, L3/R3 on two-stick models.

## Why

RetroArch's hotkey binds are raw joypad indices, and this one pad is numbered three ways: the universal cfg is written in udev numbering, the 64-bit build reads the pad through the staged SDL2 (three higher, stick clicks interleaved with the triggers), and the 32-bit build through linuxraw (udev buttons, d-pad as axes). The previous translation covered six binds for sdl2 only. Consequences on Development:

- With the shipped `Custom` option the modifier stayed at udev `6`, which is **X** under sdl2 while the menu toggle was also moved to X: holding X for five frames froze input and opened the menu on release (SPR-MED-094).
- Exit was on **A** (index 0 is A on this pad, B everywhere else); screenshot, slow-motion, shader and the state slots were unbound; the README's layout did not apply.
- The 32-bit build had fast-forward on **L2** on stick models (shipped index assumed the stickless layout).
- The RG40XX V bound R3 to the MENU tap pulse (`KEY_GOTO`) and advertised a right stick it lacks.
- PPSSPP used the label-named map, so Cross sat on the right-hand A button (every other device has it on the bottom button), and three of the four platforms had no `controls-*.ini` at all.
- DraStic, mupen64plus, PCSX-ReARMed, Flycast, YabaSanshiro, ScummVM, BigPEmu and ffplay either had 360-keyed configs that never matched this pad or never received an SDL map.

## What changed

One commit per layer:

1. **`AnbernicXXCommon.cfg`** classifies `BASEOS_TARGET` into `XX_PAD_LAYOUT` (2stick / 1stick / nostick) once; SDL maps `_XX_GC_2STICK/_1STICK/_NOSTICK`; `RA_HOME_VAL` corrected to MENU (8). Retires the three hand-copied model lists (SPR-LOW-050).
2. **`ra_functions.sh`**: `apply_xx_hotkeys_from_autoconfig` replaces `apply_xx_hotkey_for_driver`. Every hotkey is read by name from the autoconfig RetroArch is about to load, on the sdl2 and linuxraw paths (sed script file, busybox-safe). `Custom` translates only spruce's own udev literals and keeps a modifier the user set inside RetroArch. Autoconfig writers get a one-stick arm. Universal cfg carries the fleet layout in udev numbering.
3. **PPSSPP**: positional map; `controls-`/`ppsspp-` ini for RG28XX, XX640480, XX720480.
4. **DraStic**: `drastic-<platform>.cfg` (+ `.bak`) for all four platforms in the staged SDL2 numbering, touch cursor on the right stick, press on L3.
5. **mupen64plus**: `[Linux: ANBERNIC-keys]` in `InputAutoCfg.ini`, per-layout Z trigger, positional map for gptokeyb2, `--set CoreEvents` chords in this pad's numbering (MENU + START stop, + R1 save, + L1 load, + hat slot, + B reset, + Y screenshot, + X pause, + A mute, + SELECT speed limiter).
6. **PCSX-ReARMed / YabaSanshiro**: pad sections written per layout at launch (pcsx_64 links SDL 1.2 and falls back to joydev numbering; yabasanshiro keys on GUID).
7. **Flycast, ScummVM, BigPEmu, ffplay**: receive the SDL map (positional for flycast and the gptokeyb bridges, label-named for ScummVM).
8. **MENU as modifier**: universal cfg ships MENU; the `Menu` choice is offered to `ANBERNIC_RGXX`; writers record MENU's index in a comment line (RetroArch has no player bind for a guide button).

Unchanged by design: DSperate (already positional with MENU as modifier fleet-wide), Ports, PICO-8 base map.

## Verification

- Local harness: the translation was run for every layout × build × modifier option and yields the layout above; a user-set modifier under `Custom` survives.
- Host contract suite (`unittest/run_tests.py --category host`) before/after diff against Development: no new failures; fixes the recorded failure `test_xxcommon_sourcing_platforms_ship_ppsspp_default_configs`. The XX contract tests were rewritten for the three-layout model plus a functional hotkey test (wrapper repo).
- **Not yet run on hardware.** Device checks before merge (CubeXX for two-stick, RG SP for stickless): RetroArch chords under `Menu` and under `Custom` on a fresh cfg on both builds; PPSSPP face layout (Cross on the bottom button); DraStic MENU + touch cursor; pcsx_64 pad detection (`binddev` match) and hat axes; mupen64plus pad detection and MENU+START; yabasanshiro pad. The PCSX section rests on SDL 1.2's joydev fallback and the one-stick linuxraw axes are derived, not measured.

## Known trade-off

MENU is both spruce's tap/hold home key and the RetroArch modifier here, exactly as on the TrimUI line: `homebutton_watchdog` cancels the tap and hold when a second key arrives, so chords work, but a chord slower than `menuHoldTime` (shipped 1 s) triggers the hold action first. That is the fleet's existing behaviour, not introduced here; the design note is in the controller-matrix research doc.


CatalyticArkun:
Follow-up for the update path and the RetroArch reset tasks (two more commits).

**What the updater preserves.** `spruceBackup` keeps `retroarch-AnbernicRG_XX-universal.cfg`, `RetroArch/.retroarch/config`, `Emu/DC/config`, `Emu/SATURN/.yabasanshiro` and the whole PPSSPP tree under `Saves/.config/ppsspp`; `spruceRestore` extracts them over the new payload, then runs the version-gated upgrade scripts, then merges `spruce-config.json`.

**Per file on an existing card:**
- Universal cfg action binds: self-heal, the launcher rewrites every hotkey by name on each launch.
- Universal cfg modifier: does not self-heal under `Custom` (it keeps whatever the cfg holds), so `UpgradeScripts/4.3.7.sh` moves spruce's own old literals (6 = old shipped default, 9 = the old Select translation) to MENU (8). Any other value is a user's RetroArch choice and stays. The option is read from the backup copy of `spruce-config.json` because the merge runs after the scripts.
- PPSSPP: the three new per-platform ini pairs are seeded into the live tree when absent.
- DraStic, mupen64plus, PCSX: live in `Emu` directories the updater replaces; PCSX and YabaSanshiro pad sections are regenerated at launch.
- `spruce-config.json`: the merge keeps `selected` and takes the new options list, so "Menu" appears on existing cards.

**Reset tasks and defaults:** `resetRA.sh` restores from `retroarch-AnbernicRG_XX-universal.cfg.bak`, now updated to the new layout. `resetRAHotkeys.sh` (`set_default_ra_hotkeys`) targeted a `retroarch-<PLATFORM>.cfg` that exists for none of the XX platforms (SPR-MED-030); it now writes the fleet defaults into the universal cfg. The XX DraStic configs were added to the backup list so user edits survive future updates like every other platform's.

`4.3.7.sh` assumes the next release is 4.3.7; rename if the version plan differs. Host suite before/after diff against Development: no new failures.

PR #1660, opened 2026-09-10
Add low battery warning display and handling

pmsobrado:
Here's my attempt to fix low battery warning making the screen black when the user accepts the message, at least on Flip devices. I tested this myself on my Flip with Spruce 4.3.4 running.

Also, I made pressing A to dismiss not perform an A press action on whatever screen behing the warning message.

Maybe the code can be improved, cleaned, or maybe I touched the wrong files. All I can say is that it works for me. Feel free to change the code or close the PR and make your own fix :)

### DIFF
--- a/spruce/scripts/low_power_warning.sh
+++ b/spruce/scripts/low_power_warning.sh
@@ -71,6 +71,58 @@ init_battery_log() {
     fi
 }
 
+show_low_battery_warning() {
+    display -t "Battery has $CAPACITY% left. Charge or shutdown your device." \
+        --add-image "/mnt/SDCARD/spruce/imgs/displayAcknowledge.png" 1.0 240 middle
+
+    /mnt/SDCARD/spruce/flip/bin/python3 - <<'PY'
+import os
+import fcntl
+import struct
+
+DEVICE = "/dev/input/event5"
+A_BUTTON = 305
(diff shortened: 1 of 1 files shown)


spruce/scripts/low_power_warning.sh
```diff
@@ -71,6 +71,58 @@ init_battery_log() {
     fi
 }
 
+show_low_battery_warning() {
+    display -t "Battery has $CAPACITY% left. Charge or shutdown your device." \
+        --add-image "/mnt/SDCARD/spruce/imgs/displayAcknowledge.png" 1.0 240 middle
+
+    /mnt/SDCARD/spruce/flip/bin/python3 - <<'PY'
```
chrisj951, line 78: DEVICE_PYTHON3_PATH is what should be used

spruce/scripts/low_power_warning.sh
```diff
@@ -71,6 +71,58 @@ init_battery_log() {
...
+import os
```
chrisj951, line 79: This is also very device specific based on inputs being read / values checked. You should be able to combine display and confirm() or acknowledge() from helper functions to get what you want and not have it be device specific. Anything broken there should be fixed vs spinning off a flip specific low battery warning

PR #1662, opened 2026-09-10
Add Xroar Emulator for Tandy Color Computer support

ImCoKeMaN:
Looking to add the xroar color computer emulator, but this is not complete as it currently looks to be loading a retroarch core instead of xroar


Emu/COCO/config.json
ryanmsartor, line 1: one bit of feedback here - you should include a "Devices" key in your json to filter out the 32-bit-only spruce targets. Check out Emu/SATURN/config.json for an example.

ryanmsartor: ....that or include a build that will launch on those devices (A30 and Miyoo Mini family)

ImCoKeMaN: updated that and fixed my launch file

PR #1677, opened 2026-09-14
Cancel home button on brightness chord WITHIN PyUI.

zenkalia:


chrisj951:
There's some other code creating problems that this is just circumventing but not addressing. If you have time the update should be

1) Add a new key to App/PyUI/py-ui-config.json of ignoreMenuCombinationPresses
2) In Controller.py if ignoreMenuCombinationPresses is true, then have check_for_hotkey skip the actual performance of the hotkey, but report back that it WAS a hotkey press

This should prevent the rest of the code from seeing the menu, and allow delegating the hotkey solely to outside of PyUIs control instead of it being a weird shared handoff

chrisj951 (changes requested):
See comments

chrisj951:
Closing as an alternate implementation has be done

zenkalia:
🤔  Do you envision ignoreMenuCombinationPresses getting configured anywhere? I don't think I understand how this option would work. Would it ever be false?

Also, pursuing that implementation is causing some kind of race condition that isn't present in my version utilizing `menubtn_cancelled` ... and i really can't figure out why it's happening. 😮‍💨

PR #1682, opened 2026-09-14
Limited scope PyUI fixes

Sundownersport:
**RG28XX: supports_wifi on a live USB radio check**
69494f72c removed the /tmp flag constants `supports_wifi()` still read, so the top bar raised AttributeError on every render. Per review, `supports_wifi()` now returns `self._usb_radio_present()`: the live sysfs check against `WIFI_USB_IDS` (or the class's `USB_IDS` fallback) stays, and the /tmp flag reads are gone. The shell still owns the module-failed case (`device_has_wifi_radio`; `wifi.sh` records `no_radio`).

**TrimUI: drop call to removed _prepare_for_power_action in reboot**
Makes `reboot()` match `power_off()`. 69494f72c removed this call from `power_off()` and deleted the method, but left the call in `reboot()`. Affects all four TrimUI devices, since none override `reboot()`.

**Miniloong: accept include_wifi in startup_init**
`mainui.py:216` calls `startup_init(include_wifi=False)` in `-startupInitOnly` mode, which raised `TypeError` on Miniloong. The signature now matches the base class and every other device; the constructor's no-argument call still works via the default, and Miniloong's init doesn't touch WiFi, so behaviour is unchanged.

**Main menu popup: call Theme.popup_menu_rows**
Was passing the bound method instead of its return value. Now matches the other five popups and its own `cols` line.

**Game system sort setting: only open picker on A**
`ControllerInput.A` on its own is always truthy, so L1/R1 on this row also opened the sort picker. It now compares against `input` like the two branches above.

**Drop stale WIFI_USB_IDS comments and the unused dongle export**
The Flip, Brick, Brick Pro and Smart Pro cfg comments said `usb_wifi_dongle.sh` exports `WIFI_USB_IDS` for PyUI. Nothing on those devices reads it, so those comments and the export are removed. The RG28XX's own `WIFI_USB_IDS` in `AnbernicRG28XX.cfg` is unchanged and still inherited by PyUI.


App/PyUI/main-ui/devices/anbernic/anbernic_rg28xx.py
```diff
@@ -15,6 +15,8 @@ class AnbernicRG28xx(AnbernicXXCommon):
     spruce's wifi.sh, whose enable_wifi is the only path that loads the USB driver.
     """
 
+    WIFI_UNAVAILABLE_FLAG = "/tmp/wifi_unavailable"
+    WIFI_RADIO_ABSENT_FLAG = "/tmp/wifi_radio_absent"
```
chrisj951: See if it can do something via startup arg or a cfg file instead of a tmp file

chrisj951 (approved):
Left one comment but good to merge anyways for now

PR #1686, opened 2026-09-15
properly prevent the popup menu from appearing when you chord menu + volume to change the brightness.

zenkalia:
not sure how much better this implementation is since it still relies on `menubtn_cancelled` but it's most respectful of separated concerns.

:shrug:

attempt #2 after https://github.com/spruceUI/spruceOS/pull/1677

some more chatter here: https://discord.com/channels/1252066540874891355/1252066540874891358/1549270034600169562


chrisj951:
I still don't like that this can have a race condition. The change that is needed is to have PyUI monitor the volume buttons but not send vol up/down signals

chrisj951:
Should be fixed via another method on latest Development branch

Issue #1489, opened 2026-04-20
Miyoo Mini Plus - time sync on boot w/ WIFI?


fredwu:
Hi,

I have an older MM+ without RTC, SpruceOS does not seem to update/sync the time on boot. I noticed that RTC time sync was implemented from way back (see #84), so I'm not sure whether this is due to the implementation not being carried across for MM+ yet, or that older MM+ do not have built-in RTC.

FYI OnionOS detects whether the device has RTC or not (see https://github.com/OnionUI/Onion/pull/1665).

I'm hoping SpruceOS would have something similar. 🙏 

Thanks for the great work!

P.S. Was on 4.1.1, just updated to 4.2.0.

PR #1688, opened 2026-09-15
Fix time sync timeout syntax on legacy BusyBox

brendannoble:
## Problem

On a Miyoo Mini Plus running v4.4.0, automatic clock repair fails even though manual NTP works. The device ships BusyBox v1.20.2, whose `timeout` syntax is `timeout -t SECS COMMAND`. The shared helper instead invokes `timeout SECS COMMAND`, producing `timeout: can't execute '20': No such file or directory` (exit 127). Both NTP and HTTP fallback fail before starting, and their stderr redirection hides the cause.

## Change

Probe modern timeout syntax with `true`; use the legacy `-t` form when the probe fails. The requested operation runs only once, retains its arguments and exit status, and keeps the existing direct-execution fallback when timeout is absent. This fixes the shared path used by clock repair and timezone HTTP requests.

Related to #1489.

## Validation

- Rebased and checked against `Development` at `83cd7dffad038361ceab67f2da990dd9a43284e9`. Its original helper reproduced the legacy failure (exit 127, zero backend calls); the rebased fix passed all three ad hoc interface checks, real GNU timeout checks, shell syntax validation, and whitespace checks. An additional agent review found no blocking issues in the Development diff.
- Ad hoc checks covered modern, legacy, and missing timeout, including spaced/empty arguments, child failure status, and exactly one backend execution. The legacy case failed against the original implementation with exit 127 and no backend invocation, then passed with this change. These checks are not added to the repository.
- Shell syntax checks and `git diff --check` passed.
- GNU coreutils timeout 8.32: child exit 7 preserved; five-second sleep terminated after one second (exit 124).
- Physical Miyoo Mini Plus, BusyBox v1.20.2: candidate helper tested in a temporary SSH shell; child exit 7 preserved; five-second sleep terminated after one second (exit 143). The installed HTTP header helper, using this candidate wrapper, successfully returned the time server's Date header.
- Two independent agent reviews completed with no blocking findings.

No firmware files were changed during device checks. A complete cold boot with the patched firmware has not been tested; modern BusyBox was modeled in the ad hoc checks but not tested on a separate physical device.


Sundownersport:
Please target and test changes against the Development branch.

brendannoble:
@Sundownersport Rebased onto Development, updated the PR target, and verified the fix against that branch.

Sundownersport:
Ok thanks, I’ll take a look later tonight

brendannoble:
> Ok thanks, I’ll take a look later tonight

Thanks, I appreciate it!

PR #1695, opened 2026-09-17
Add MagicX A133P device support

ryanmsartor:
WIP


App/PyUI/launch.sh
```diff
@@ -339,14 +339,20 @@ case "$PLATFORM" in
     ;;
 
 ############################################################
-# MagicX Zero28
+# MagicX A133P family (Mini Zero 28, Zero 40, XU20 V32)
```
ryanmsartor, line 342: A little unclear to me why this would be a joined case, only to immediately split the case again. Is there any particular rationale there?

ryanmsartor: Well, I guess there _is_ some shared code. Eh, if it works it works.

App/PyUI/main-ui/devices/magicx/magicx-system.json
ryanmsartor, line 1: Worth discussing with the team whether we should go with a shared system config or split.

App/PyUI/main-ui/display/display.py
```diff
@@ -244,7 +244,7 @@ def _init_display(cls):
                 width, height = Device.get_device().screen_width(), Device.get_device().screen_height()
             else:
                 width, height = display_mode.w, display_mode.h
-                #PyUiLogger.get_logger().info(f"Display size: {width}x{height}")
+                PyUiLogger.get_logger().info(f"Display size: {width}x{height} (device says {Device.get_device().screen_width()}x{Device.get_device().screen_height()} rotation {Device.get_device().screen_rotation()})")
```
ryanmsartor: Is this just turned on for testing? May wanna disable once in production?

Emu/COCO/config.json
```diff
@@ -1,40 +1,42 @@
...
     "MIYOO_FLIP",
     "GKD_PIXEL2",
-    "ANBERNIC_RG34XXSP","ANBERNIC_RG28XX",
-    "ANBERNIC_RGXX640480"
-    ],
-    "scaling_min_freq": 1008000,
-    "alternativeFolderNames": [
-        "coco"
-    ]
+    "ANBERNIC_RG34XXSP",
```
ryanmsartor: Not really related to your MagicX changes, but I think these device names in the emu configs need to be updated for the XX family.

spruce/scripts/helperFunctions.sh
```diff
@@ -108,7 +115,9 @@ device_names() {
         Pixel2)           echo "GKD_PIXEL2" ;;
         RGB30)            echo "RGB30" ;;
         Miniloong)        echo "MINILOONG_POCKET1" ;;
-        Zero28)           echo "MAGICX_ZERO28" ;;
+        Zero28)           echo "MAGICX_ZERO28"; echo "MAGICX_A133P" ;;
```
ryanmsartor, line 118: Double echoes? How's this pan out in practice? Does the second mask the first?

spruce/scripts/platform/Zero40.cfg
ryanmsartor, line 1: May be worth considering a shared platform config for the 3x MagicX A133Ps, with specific overrides added in breakout configs? See XX family for example.

ryanmsartor:
I glossed over most of the PyUI changes, but my overall thought is if we test this and make sure none of the existing devices break, this is probably good to merge soon.

CatalyticArkun:
Converting to a draft to prevent accidental merging

App/PyUI/main-ui/devices/magicx/magicx_a133p_device.py
```diff
@@ -0,0 +1,191 @@
...
+
+        script_dir = Path(__file__).resolve().parent
+        self._load_system_config(system_json_path, script_dir / 'magicx-system.json')
+        if main_ui_mode:
+            self.miyoo_games_file_parser = MiyooGamesFileParser()
+            threading.Thread(target=self.startup_init, daemon=True).start()
+            self.config_watcher_thread, self.config_watcher_thread_stop_event = FileWatcher().start_file_watcher(
+                system_json_path, self.on_system_config_changed, interval=0.2,
+                repeat_trigger_for_mtime_granularity_issues=True)
+            if PyUiConfig.enable_button_watchers():
```
chrisj951: This is old, the enable button watchers should be on the button processing, not the listener. We always want all keys to fire events, we just choose to ignore them

App/PyUI/main-ui/themes/theme_patcher.py
```diff
@@ -145,7 +146,8 @@ def scale_theme(cls, config_path, theme_width, theme_height, target_width, targe
                      scale,
                      width_multiplier,
                      height_multiplier,
-                     aspect_ratio_reset)
+                     aspect_ratio_reset,
```
chrisj951: Can we pull this out and have it separate in its own PR with more discussion? Will be easier to merge if handled separately

App/PyUI/main-ui/views/grid_view.py
```diff
@@ -319,9 +319,39 @@ def get_selection(self, select_controller_inputs=[ControllerInput.A]):
...
 
+    def cell_at_touch_point(self, point):
+        """Option index under a logical-space tap, or None (mirrors _render_cell)."""
+        if point is None or not self.options:
+            return None
+        x, y = point
+        x_index = int((x - self.x_pad) // self.icon_width) if self.icon_width else 0
+        if x_index < 0 or x_index >= self.cols:
+            return None
+        top = Display.get_top_bar_height(False)
```
chrisj951: Probably safer to not set False and allow the default for now. I think we need to not put force_include_top_bar and instead track in display if the last draw had the top bar

App/PyUI/main-ui/views/grid_view.py
```diff
@@ -319,9 +319,39 @@ def get_selection(self, select_controller_inputs=[ControllerInput.A]):
...
+                if hit is not None:
+                    if hit == self.selected:
+                        return Selection(self.get_selected_option(), ControllerInput.A, self.selected)
+                    self.selected = hit
+                    self.correct_selected_for_off_list()
 
         return Selection(self.get_selected_option(), None, self.selected)
 
+    def cell_at_touch_point(self, point):
+        """Option index under a logical-space tap, or None (mirrors _render_cell)."""
```
chrisj951: Too tired to logic this out so going to just assume it's right if it was tested. Left one comment on the default

App/PyUI/main-ui/views/list_view.py
```diff
@@ -86,8 +86,45 @@ def get_selection(self, select_controller_inputs = [ControllerInput.A]):
...
+                if hit is not None:
+                    if hit == self.selected:
+                        self.selection_made()
+                        return Selection(self.get_selected_option(), ControllerInput.A, self.selected)
+                    self.selected = hit
 
         return Selection(self.get_selected_option(), None, self.selected)
+
+    def row_at_touch_point(self, point):
+        """Index of the visible row under a logical-space tap, or None.
```
chrisj951: This is fine for a first implementation but 'x' might want to actually be taken into consideration.

If you're on an image list view and you tap the image you would probably expect it to launch the game and NOT the row that would be behind the image.

Not a big deal as that is kind of niche.

App/PyUI/main-ui/views/list_view.py
```diff
@@ -86,8 +86,45 @@ def get_selection(self, select_controller_inputs = [ControllerInput.A]):
...
+
+        Rows are laid out from base_y_offset in line_height steps (the
+        NonDescriptiveListView family); other lists share the window evenly.
+        """
+        if point is None or not self.options:
```
chrisj951, line 106: Im not sure all this logic works for all subclasses, as long as it's been tested that should be fine though

App/PyUI/main-ui/views/list_view.py
```diff
@@ -86,8 +86,45 @@ def get_selection(self, select_controller_inputs = [ControllerInput.A]):
...
+            return None
+        x, y = point
+        visible = self.current_bottom - self.current_top
+        if visible <= 0:
+            return None
+        top = getattr(self, 'base_y_offset', None)
```
chrisj951: I do not like the use of getattr here. It is assuming the child classes HAVE these variables. If subclasses must have them then make an @abstractmethod and have each subclass implement it

spruce/scripts/platform/device_functions/trimui_a133p.sh
```diff
@@ -334,7 +334,8 @@ device_needs_strict_unmount() {
 # dongle whose module will not load, means the onboard radio exactly as before.
 #
 # Defined only for a cfg that opted in (the platform cfg is sourced before
-# this file). This file is also sourced by RGB30.sh and Zero28.sh for its
+# this file). This file is also sourced by RGB30.sh and by the MagicX A133P
```
chrisj951: Can we break this file into a133p and trimui_a133p. There are trimui SPECIFIC things in here and the name will be confusing to keep track of.

a133p sourced by trimui_a133p, sourced by trimui_<device>

a133p sourced by magicx_a133p

spruce/scripts/helperFunctions.sh
chrisj951: This also confuses me unless it is checking for <x> in string potentially

PR #1697, opened 2026-09-18
Allow limiting certain about settings to certain devices

chrisj951:
Also add a way to get the baseOS version


ryanmsartor:
Beautiful, love the simplicity, but won't this still leave an "unknown" firmware version readout for the XXes?

chrisj951:
> Beautiful, love the simplicity, but won't this still leave an "unknown" firmware version readout for the XXes?

No, the entry is limited to just the XX devices. (Probably need to add magicx there in the future)

PR #1698, opened 2026-09-18
Fix idle shutdown on TSPS

zenkalia:
TSPS was never falling asleep due to stick jitter and rumble keepalive messages. Change should be constrained to just TSPS but what do I know?


Sundownersport:
@zenkalia just want to be sure that you tried calibrating your sticks to see if that fixed this issue before making these changes?

Sundownersport:
I tested on a Smart Pro S, and the cause turns out to be different from what the PR describes.

  On the TSPS, set_event_arg_for_idlemon never sets EVENT_ARG, so idlemon runs with no -e. Its built-in default is /dev/input/event3, which on the TSPS is audiocodec Headphones, the headphone jack switch, not the
  pad (event4, TRIMUI Player1). idlemon never sees stick jitter, rumble, or any real input. I confirmed it by checking /proc/<pid>/fd on the running idlemon.

  What that looks like on a TSPS with "Shutdown while idle in menu" at 2m:
  - Untouched: shuts down at ~2 min, as expected.
  - Held the stick the whole time (PyUI saw it, so the screensaver never came on): still shut down at ~2 min.

  So idle shutdown fires even while the device is in use. The gate script would hide that, because it restarts idlemon on real input, but idlemon would still be reading the headphone jack.

  The fix is a single line in SmartProS.sh, the same one Miniloong and the Pixel2 use:

  set_event_arg_for_idlemon() {
      EVENT_ARG="-e $EVENT_PATH_READ_INPUTS_SPRUCE"
  }
  
  This also proved to be an issue on the MagicX Zero28. commit is here https://github.com/spruceUI/spruceOS/commit/4c8e99fd96058b1652f1a1dd1490206a7bfde0a6

PR #1701, opened 2026-09-19
Fixed DSi release year

Arkhaeon:
DSi release year was 2008 in Japan, incorrectly put as 2004 in config.json.
Found through incorrect system sorting behaviour.


CatalyticArkun:
Good catch, trivial fix 👍

